PR #35, opened 2023-11-17
Bug fixed: added UTC Timezone to all datetimes. Fix #12

### LINKED ISSUE #12: Malformed date from apps
TL;DR:

We need to change this kind of datetime constructors:

https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/c3ef283da54cd6afec9d38de75b67cb36cbdf4fd/lib/utils/Utils.dart#L188

From 

DateTime.now().toIso8601String()

to 

DateTime.now().toUtc().toIso8601String()

The first format DateTime.now().toIso8601String() is **NOT UTC**, and it should be. When this kind of date is uploaded to the server, the server assumes this is UTC, so it adds two hours to the time.

You can play around with this at this url: [https://dartpad.dev](https://dartpad.dev

I'm writing this issue at around 16'11 local time. If I execute this code:

```
void main() {    
  print(DateTime.now().toIso8601String());
  print(DateTime.now().toUtc());
  print(DateTime.now().toUtc().toIso8601String());
}
```

It gives as a result
```
2023-06-08T16:11:51.503
2023-06-08 14:11:51.504Z
2023-06-08T14:11:51.504Z
```
We are in UTC zone +2 (+1 for zone, +1 for summer time) so the correct times are the second and third lines. Since dribba is uploading the time in the first format, the server accepts it and assumes it's utc-correct (it's not), so it adds +2 to the time.

As a result, the timestamps for phone_upload_time, creation_time and version_time are incorrect (most of the times are 2 hours in the future of the correct time).

epou:
```bash
find . -name '*.dart' -exec sed -i 's/DateTime.now()/DateTime.now().toUtc()/g' {} \;
```

### DIFF
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -522,7 +522,7 @@ class ApiSingleton {
         'masked_lat': lat,
         'masked_lon': lon,
         'power': power,
-        'phone_upload_time': DateTime.now().toIso8601String(),
+        'phone_upload_time': DateTime.now().toUtc().toIso8601String(),
       };
 
       final response = await http
--- a/lib/pages/forms_pages/adult_report_page.dart
+++ b/lib/pages/forms_pages/adult_report_page.dart
@@ -262,7 +262,7 @@ class _AdultReportPageState extends State<AdultReportPage> {
         if (Utils.savedAdultReport!.country != null) {
           List<Campaign> campaignsList =
               await (ApiSingleton().getCampaigns(Utils.savedAdultReport!.country) as FutureOr<List<Campaign>>);
-          var now = DateTime.now();
+          var now = DateTime.now().toUtc();
           if (campaignsList.any((element) =>
               DateTime.parse(element.startDate!).isBefore(now) &&
               DateTime.parse(element.endDate!).isAfter(now))) {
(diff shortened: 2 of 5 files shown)


pablogalve (approved):
Code change looks good to me according to bug description in #12
This can be tested together with the other changes for the next release 👍 
I have added this flag to close the issue automatically once this is merged
image

aescobarr:
One important thing about this: when the mobile app posts a report, it sets a field called package_version. For all the releases performed by Dribba, this package_version is 32. It would be necessary to change this number to 33 or whatever, because this number helps us group very nicely the reports affected by this bug (i.e. all reports with package number 32). When we roll out the next update (33) the adoption on user's phones will be gradual, so there will be reports with wrong dates entering the system for a while. We could set up a script which checks (and corrects) the wrong dates in reports with package.version == 32.

So please @pablogalve @JohnPalmer remember this the next time you perform a release which includes this fix.

A.

aescobarr (approved):

JohnPalmer:
I agree with Agusti's point - this is really important

pablogalve:
@aescobarr thank you, I'll have this into account. Is the package_version value of 32 coming from anywhere or is it an arbitrary value? 

If the second, we could consider instead to use the app version for consistency (currently 1709)

@epou Automations are our friend, thanks for providing some examples

aescobarr:
@pablogalve as far as I know, the numbers are  consecutive. If I look at the database, the package_version is a sequence from 0 to 32. I don't think it matters very much as long as we have different numbers for different releases; the next logical number to assign would be 33 but there is no problem if it's 34 or 35.

epou:
Hey team, I think we are ready to merge this PR. Appreciate your reviews!

Regarding the discussions about the app release version number, and in order to keep things organized, I've created a dedicated GitHub issue for this matter. So let's continue the discussion there #36

PR #73, opened 2024-01-30
CI: Analyze for warnings and errors, ignoring "info" level messages

DigitalSeneca:
Took the base workflow from [this branch](https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/tree/3-set-up-unit-testing-and-ci-workflow) that was having the problem of detecting "info" as error and failing the build:
image

I modified:
- Use flutter 3.0.0
- Use grep to filter and only fail the PR with warnings and errors

### DIFF
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -18,9 +18,6 @@ jobs:
         channel: 'stable'
         flutter-version: '3.0.0'   
 
-    - name: Get flutter dependencies
-      run: flutter pub get
-
     - name: Statically analyze the Dart code for any errors/warnings, ignoring "info" level warnings
       run: |
         if flutter analyze 2>&1 | grep -q -E 'error:|warning:'; 
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -26,7 +26,8 @@ jobs:
       run: flutter pub get
 
     - name: Statically analyze the Dart code for any errors, ignoring "info" level warnings
-      run: flutter analyze 2>&1 | grep -q -E 'error:|warning:'; 
+      run: |
+        if flutter analyze 2>&1 | grep -q -E 'error:|warning:'; 
         then exit 1
         else exit 0
         fi
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -8,11 +8,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Set up Dart SDK
-      uses: dart-lang/setup-dart@v1
-      with:
-        dart-sdk: '2.17.6'
-
     - uses: subosito/flutter-action@v2
       with:
         channel: 'stable'
--- a/.github/workflows/static_analysis.yml
+++ b/.github/workflows/static_analysis.yml
@@ -6,9 +6,9 @@ jobs:
   static_analysis:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,33 @@
+name: Static analysis
+on:
+  pull_request:
+
+jobs:
+  static_analysis:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
```
epou, line 9: Could you try upgrading to `actions/checkout@v4`? (see: https://github.com/actions/checkout)

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    - uses: actions/setup-java@v1
+      with:
+        java-version: '12.x'
```
epou, line 13: Why is java needed?

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    - name: Set up Dart SDK
+      uses: dart-lang/setup-dart@v1
+      with:
+        dart-sdk: '2.17.6'
+
+    - uses: subosito/flutter-action@v1
```
epou, line 20: Could you try upgrading to `subosito/flutter-action@v2`? (see: https://github.com/subosito/flutter-action)

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,33 @@
...
+      with:
+        channel: 'stable'
+        flutter-version: '3.0.0'   
+
+    - name: Get flutter dependencies
+      run: flutter pub get
+
+    - name: Statically analyze the Dart code for any errors/warnings, ignoring "info" level warnings
+      run: |
+        if flutter analyze 2>&1 | grep -q -E 'error:|warning:';
```
epou, line 30: Could you try adding:  `--no-pub` (`flutter analyze --no-pub`) to skip running `pub get` again?

epou (approved):

.github/workflows/static_analysis.yml
DigitalSeneca: seems like it's not needed, I removed it

.github/workflows/static_analysis.yml
DigitalSeneca: I have upgraded both `actions/checkout@v4` and `subosito/flutter-action@v2` to the latest version instead of previous versions

.github/workflows/static_analysis.yml
```diff
@@ -0,0 +1,33 @@
...
+    - uses: actions/checkout@v1
+
+    - uses: actions/setup-java@v1
+      with:
+        java-version: '12.x'
+
+    - name: Set up Dart SDK
+      uses: dart-lang/setup-dart@v1
+      with:
+        dart-sdk: '2.17.6'
```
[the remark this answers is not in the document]
DigitalSeneca: actually not, I removed it. By specifying `flutter-version: '3.0.0'`, there's no more need to specify dart anymore

.github/workflows/static_analysis.yml
epou: nice, my thoughts were that it'd be only needed for building the Android APK, but only for analyzing the code it would be unnecessary.... just to speed up things as much as possible

DigitalSeneca:
@epou Thanks for the review, it was much longer than needed

Issue #69, opened 2024-01-29
Automatically add new translations from localise.biz using Github Actions


epou:
See: https://github.com/marketplace/actions/sync-translations-from-localise-biz

Ideally, it should open a new pull-request when detecting modification on the translations.
See: https://github.com/marketplace/actions/create-pull-request
(
PR_TITLE="Update Translations"
PR_BRANCH="update-translations-$(date +%s)"
PR_BODY="Automated pull request to update translations."
)

DigitalSeneca:
What should be the target branch for those pull requests? Right now it's v33, but then it will be changing every x amount of time to v34, v35...
Should it be 'main'? It's always possible to manually update the target, but the more automation the better

epou:
Yes, using 'main' is wiser for this kind of issue related to the repository workflow, and that does not add any change to the app itself.

PR #74, opened 2024-01-30
Change title for report type on "My data" -> List

### LINKED ISSUE #63: Change title for report type on "My data" -> List
Change the message:
```
Notification of the day DD-MM-YYYY
Approximate location {LOCATION}
At: HH:mm
```
For
```
{Report.type}
Approximate location {LOCATION}
At: DD-MM-YYYY- HH:mm
```

Caution with translations!

DigitalSeneca:
fixes #63 

| English | Spanish |
| image | image |
| image | - |

### DIFF
--- a/assets/language/en.json
+++ b/assets/language/en.json
@@ -322,5 +322,8 @@
     "gl_ES": "Galician",
     "eu_ES": "Basque",
     "sv_SE": "Swedish",
-    "bn_BD": "Bengali"
+    "bn_BD": "Bengali",
+    "single_bite": "Bite",
+    "single_mosquito": "Mosquito",
+    "single_breeding_site": "Breeding site"
 }
--- a/assets/language/en.json
+++ b/assets/language/en.json
@@ -164,7 +164,7 @@
     "map_txt": "Map",
     "list_txt": "List",
     "report_of_the_day_txt": "Notification of the day ",
-    "location_txt": "Approximate location:",
+    "location_txt": "Approximate location",
     "at_time_txt": "At:",
     "registered_location_txt": "Registered location",
     "exact_time_register_txt": "Time of report",
--- a/assets/language/en.json
+++ b/assets/language/en.json
@@ -164,7 +164,7 @@
     "map_txt": "Map",
     "list_txt": "List",
     "report_of_the_day_txt": "Notification of the day ",
-    "location_txt": "Approximate location",
+    "location_txt": "Approximate location:",
     "at_time_txt": "At:",
     "registered_location_txt": "Registered location",
     "exact_time_register_txt": "Time of report",
--- a/lib/pages/my_reports_pages/components/reports_list_widget.dart
+++ b/lib/pages/my_reports_pages/components/reports_list_widget.dart
@@ -61,7 +61,7 @@ class ReportsList extends StatelessWidget {
                                   Utils.getTranslatedReportType(context, reports[index].type),
                                   fontSize: 14),
                               Style.body(
-                                  '${MyLocalizations.of(context, "location_txt")} ${reports[index].displayCity ?? ''}'),
+                                  '${MyLocalizations.of(context, "location_txt")}: ${reports[index].displayCity ?? ''}'),
                               Style.body(
                                 MyLocalizations.of(context, 'at_time_txt')! +
                                 ' ' +
--- a/lib/pages/my_reports_pages/components/reports_list_widget.dart
+++ b/lib/pages/my_reports_pages/components/reports_list_widget.dart
@@ -57,22 +57,23 @@ class ReportsList extends StatelessWidget {
                             crossAxisAlignment: CrossAxisAlignment.start,
                             children: <Widget>[
                               Style.titleMedium(
-                                  MyLocalizations.of(
-                                          context, 'report_of_the_day_txt')! +
-                                      DateFormat('dd-MM-yyyy')
-                                          .format(DateTime.parse(
-                                              reports[index].creation_time!))
-                                          .toString(),
-                                  fontSize: 14),
+                                  "TODO: report type",                                      
+                                  fontSize: 14),                                  
                               Style.body(
                                   '${MyLocalizations.of(context, "location_txt")} ${reports[index].displayCity ?? ''}'),
                               Style.body(
-                                  MyLocalizations.of(context, 'at_time_txt')! +
-                                      DateFormat.Hm()
-                                          .format(DateTime.parse(
-                                              reports[index].creation_time!))
-                                          .toString(),
-                                  color: Colors.grey),
+                                MyLocalizations.of(context, 'at_time_txt')! +
+                                ' ' +
+                                DateFormat('dd-MM-yyyy')
+                                  .format(DateTime.parse(
+                                      reports[index].creation_time!))
+                                  .toString() +
+                                DateFormat.Hm()
+                                  .format(DateTime.parse(
+                                      reports[index].creation_time!))
+                                  .toString(),
+                                color: Colors.grey
+                              ),
                             ],
                           ),
                         ),
--- a/lib/pages/my_reports_pages/components/reports_list_widget.dart
+++ b/lib/pages/my_reports_pages/components/reports_list_widget.dart
@@ -14,20 +14,20 @@ class ReportsList extends StatelessWidget {
 
   String getTranslatedReportType(BuildContext context, String? reportType){
     var translationString;
-switch (reportType) {  
-    case 'adult':  
-      translationString = 'single_mosquito';  
-      break;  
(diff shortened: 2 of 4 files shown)


lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -12,6 +12,21 @@ class ReportsList extends StatelessWidget {
 
   ReportsList(this.reports, this.onTap);
 
+  String getTranslatedReportType(BuildContext context, String? reportType){
+    var translationString = '';
+    if(reportType == 'adult'){
+      translationString = 'single_mosquito';
+    }else if(reportType == 'bite'){
+      translationString = 'single_bite';
+    }else if(reportType == 'bite'){
```
epou, line 21: 'bite' already considered. Change to "breeding_site"

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -12,6 +12,21 @@ class ReportsList extends StatelessWidget {
...
+      translationString = 'single_breeding_site';
+    }else{
+      print('Unhandled report type: $reportType');
+    }
```
epou, line 25: Maybe using a switch - case would improve readability....

```dart
switch (reportType) {
    case 'adult':
      translationString = 'single_mosquito';
      break;
    case 'bite':
      translationString = 'single_bite';
      break;
    case 'breeding_site':
      translationString = 'single_breeding_site';
      break;
    default:
      print('Unhandled report type: $reportType');
      return reportType ?? '';
  }
```

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -12,6 +12,21 @@ class ReportsList extends StatelessWidget {
 
   ReportsList(this.reports, this.onTap);
 
+  String getTranslatedReportType(BuildContext context, String? reportType){
+    var translationString = '';
```
epou, line 16: `String translationString;`

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -12,6 +12,21 @@ class ReportsList extends StatelessWidget {
 
   ReportsList(this.reports, this.onTap);
 
+  String getTranslatedReportType(BuildContext context, String? reportType){
```
epou, line 15: Does `reportType` can be null in the code?

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -57,22 +72,24 @@ class ReportsList extends StatelessWidget {
...
                               Style.body(
-                                  MyLocalizations.of(context, 'at_time_txt')! +
-                                      DateFormat.Hm()
-                                          .format(DateTime.parse(
-                                              reports[index].creation_time!))
-                                          .toString(),
-                                  color: Colors.grey),
+                                MyLocalizations.of(context, 'at_time_txt')! +
+                                ' ' +
+                                DateFormat('dd-MM-yyyy')
```
epou, line 82: Use `dd-MM-yyyy HH:mm`

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -57,22 +72,24 @@ class ReportsList extends StatelessWidget {
...
+                                  .format(DateTime.parse(
+                                      reports[index].creation_time!))
+                                  .toString() +
+                                ' ' +
+                                DateFormat.Hm()
+                                  .format(DateTime.parse(
+                                      reports[index].creation_time!))
+                                  .toString(),
```
epou, line 90: Remove this after adding `dd-MM-yyyy HH:mm` to the previous `DateFormat`

epou (changes requested):
Some improvements are suggested.

Not sure if when clicking a report on the list, the title of the 'detailed report view' is also changed to {Report.type} or not :thinking:

lib/pages/my_reports_pages/components/reports_list_widget.dart
DigitalSeneca: Right now it can theoretically be null, as defined in the Reports class. That's because after flutter 3.0.0 all types require null safety.

Maybe it's better to change the nomenclature to this so that it can't be null, but if we decide to do this, then we have to handle it correctly for this variable and also the others which are nullable right now
```
class Report {
  // other vars
  late String type;  // before: String? type;

  Report(
      // other vars
      required this.type, // before: this.type,
```

DigitalSeneca:
@epou As the screenshot shows, right now the type only works for English and is empty for other languages. This will be fixed just by adding the strings with #69 

"reports[index].displayCity" seems to be null in some cases. This was already happening before. How should this case be handled?

epou:
Could you implement a fallback to the English version when no translation is available in the specified language? I prefer seeing an English word instead of an empty value. 

Regarding `displayCity`, there's an issue reported (refer to issue #77).

DigitalSeneca:
> Some improvements are suggested.
> 
> Not sure if when clicking a report on the list, the title of the 'detailed report view' is also changed to {Report.type} or not 🤔

I have updated it, when clicking a report, it also shows {Report.type} as title instead of "Notification of the day"

> Regarding displayCity, there's an issue reported (refer to issue https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/issues/77).

Okey, I didn't see it. I'll handle it in a separate pull request

assets/language/en.json
```diff
@@ -164,7 +164,7 @@
     "map_txt": "Map",
     "list_txt": "List",
     "report_of_the_day_txt": "Notification of the day ",
-    "location_txt": "Approximate location",
+    "location_txt": "Approximate location:",
```
epou, line 167: Remove `:`

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -57,22 +58,19 @@ class ReportsList extends StatelessWidget {
...
-                                  MyLocalizations.of(
-                                          context, 'report_of_the_day_txt')! +
-                                      DateFormat('dd-MM-yyyy')
-                                          .format(DateTime.parse(
-                                              reports[index].creation_time!))
-                                          .toString(),
+                                  Utils.getTranslatedReportType(context, reports[index].type),
                                   fontSize: 14),
                               Style.body(
                                   '${MyLocalizations.of(context, "location_txt")} ${reports[index].displayCity ?? ''}'),
```
epou, line 64: Add `:` after the location_txt translated string

epou (changes requested):
Just change where the `:` is

assets/language/en.json
DigitalSeneca: okey, I'll trust the provided translations. I was just confused that some languages have the : on this string while others don't

lib/pages/my_reports_pages/components/reports_list_widget.dart
DigitalSeneca: should I remove the `:` from the strings where it's included, like ca, es, eu, gl, pt? I am afraid there would be a double `:` otherwise

lib/pages/my_reports_pages/components/reports_list_widget.dart
epou: In the backend `Report.type` is a mandatory field (see: https://github.com/Mosquito-Alert/mosquito_alert/blob/e0618655f6320c9befb29b0741a35fa096fe8c64/tigaserver_app/models.py#L416) so it would make sense to make it required in the app code too. If you think we might address this in a new issue we can create a one for checking the required variables in the code.

assets/language/en.json
epou: You're right, some translations do contain ":". I will solve that on issue #80 ensuring that these characters are on the code side, not on the translation. Keep it without the ":" for this pull request,  thought

lib/pages/my_reports_pages/components/reports_list_widget.dart
epou: I've already removed them from the localise.biz platform. So in the next export, it will be already removed :)

lib/pages/my_reports_pages/components/reports_list_widget.dart
DigitalSeneca: > If you think we might address this in a new issue we can create a one for checking the required variables in the code.

Yeah I think it makes sense to address all or most in a dedicated pull request

assets/language/en.json
DigitalSeneca: perfect

epou:
Ei! I've already given my approval, but it's still a Draft...

PR #76, opened 2024-01-31
Localize datetimes

epou:


epou:
@DigitalSeneca , any idea on how to show the datetime in the local time where the report was taken? That is, use the GPS information from the report to determine its Timezone and present the datetime in that specific local time.

DigitalSeneca:
"toLocal()" should work if the DateTime object contains timezone information
I did a bit of debugging and got the value "2024-01-26T19:27:19.182531Z" which does not indicate timezone. If there is no timezone, how can "toLocal()" know what's the expected timezone?

image

> That is, use the GPS information from the report to determine its Timezone

We can get timezone information by doing this https://stackoverflow.com/a/65581347
```
 DateTime dateTime = DateTime.now();
 print(dateTime.timeZoneName);
 print(dateTime.timeZoneOffset);
```

So we have 2 parts:
1. Create a report: It seems like data is being stored in UTC time https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/6c5993a230a86f958c8dca8bcfef2994a6b93ebe/lib/utils/Utils.dart#L363
2. Display an existing report: The API can return a date with timezone in UTC, and the mobile phone converts it to local with toLocal() or similar, by knowing the dateTime.timeZoneOffset

DigitalSeneca (commented):
"toLocal()" should work if the DateTime object contains timezone information
I did a bit of debugging and got the value "2024-01-26T19:27:19.182531Z" which does not indicate timezone. If there is no timezone, how can "toLocal()" know what's the expected timezone?

image

> That is, use the GPS information from the report to determine its Timezone

We can get timezone information by doing this https://stackoverflow.com/a/65581347
```
 DateTime dateTime = DateTime.now();
 print(dateTime.timeZoneName);
 print(dateTime.timeZoneOffset);
```

So we have 2 parts:
1. Create a report: It seems like data is being stored in UTC time https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/6c5993a230a86f958c8dca8bcfef2994a6b93ebe/lib/utils/Utils.dart#L363
2. Display an existing report: The API can return a date with timezone in UTC, and the mobile phone converts it to local with toLocal() or similar, by knowing the dateTime.timeZoneOffset

epou:
Let me check if I can expose it in the local TZ in the current API...

PR #79, opened 2024-02-01
Show string in English when there is no translation available in the local language

DigitalSeneca:
> https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/74#issuecomment-1921158145 Could you implement a fallback to the English version when no translation is available in the specified language? I prefer seeing an English word instead of an empty value.

image

### DIFF
--- a/lib/utils/MyLocalizations.dart
+++ b/lib/utils/MyLocalizations.dart
@@ -5,16 +5,15 @@ import 'package:flutter/services.dart';
 
 class MyLocalizations {
   Locale locale;
-  static Map<dynamic, dynamic>? _localisedValues;
-  static Map<dynamic, dynamic>? _englishValues;
+  static Map<dynamic, dynamic> _localisedValues = {};
+  static Map<dynamic, dynamic> _englishValues = {};
 
   MyLocalizations(this.locale) {
     locale = locale;
-    _localisedValues = null;
-    loadEnglishTranslations();
+    _loadEnglishTranslations();
   }
 
-  static Future<void> loadEnglishTranslations() async {
+  static Future<void> _loadEnglishTranslations() async {
     var jsonContent = await rootBundle.loadString('assets/language/en.json');
     _englishValues = json.decode(jsonContent);
   }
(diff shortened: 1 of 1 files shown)


lib/utils/MyLocalizations.dart
```diff
@@ -6,10 +6,17 @@ import 'package:flutter/services.dart';
...
   static Map<dynamic, dynamic>? _localisedValues;
+  static Map<dynamic, dynamic>? _englishValues;
 
   MyLocalizations(this.locale) {
     locale = locale;
     _localisedValues = null;
+    loadEnglishTranslations();
+  }
+
+  static Future<void> loadEnglishTranslations() async {
```
epou, line 17: Make it a private function since it is not intended to be used outside the class.

`_loadEnglishTranslations`

lib/utils/MyLocalizations.dart
```diff
@@ -6,10 +6,17 @@ import 'package:flutter/services.dart';
 class MyLocalizations {
   Locale locale;
   static Map<dynamic, dynamic>? _localisedValues;
+  static Map<dynamic, dynamic>? _englishValues;
```
epou, line 9: I personally think that it makes more sense to init this variables as empty dict, it simplify the code after (see translate method)

```dart
static Map<dynamic, dynamic> _localisedValues = {};
static Map<dynamic, dynamic> _englishValues = {};
```

lib/utils/MyLocalizations.dart
```diff
@@ -21,10 +28,15 @@ class MyLocalizations {
...
-        : '';
+  String translate(key) {
+    if (_localisedValues != null && _localisedValues![key] != null){
+      return _localisedValues![key];
+    }
+    if (_englishValues != null && _englishValues![key] != null){
+      // Default to English when there is not a translation available in the local language
+      return _englishValues![key];
+    }
+    return '';
```
epou, line 39: If changing `_localisedValues` and `_englishValues` from nullable to empty dict init, this code is more readable:
```dart
String translate(dynamic key) {
    return _localisedValues[key] ?? _englishValues[key] ?? '';
  }
```

DigitalSeneca: man that's nice!
But it makes more sense to be "String? key" rather than dynamic, since we know the type

epou: Ups! You're right, "String" would be the best choice!

lib/utils/MyLocalizations.dart
```diff
@@ -21,10 +27,8 @@ class MyLocalizations {
     return appTranslations;
   }
 
-  String? translate(key) {
-    return _localisedValues != null && _localisedValues![key] != null
-        ? _localisedValues![key]
-        : '';
+  String translate(String? key) {
```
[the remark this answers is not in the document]
epou: Okay :+1: , in a future we should work no that though

epou (approved):

Issue #38, opened 2023-12-15
Automate release of new versions to google play / apple store


epou:
Use Github Actions to upload new releases to google play / apple store

See:
- https://akshaaatt.medium.com/publishing-to-the-app-store-with-fastlane-and-github-actions-3c6dd80eeb4e
- https://fastlane.tools

DigitalSeneca:
Right now we can't build the appbundle on the cloud because there are some needed files that are not in the repo because they contain secrets. When I add those files locally, I can do the build locally. Do you know if it's possible to add those files as repository/actions secrets? or any other way that those files can be accessed from github actions?

.gitignore
```
android/JRBP_Keystore
android/key.properties
android/key.jks
android/app/google-services.json
```

epou:
Not sure what those keys are for... But if they are for uploading to play store take a look at: https://docs.flutter.dev/deployment/cd#fastlane

PR #81, opened 2024-02-05
Generate environment specific configuration

### LINKED ISSUE #41: Generate environment-specific configuration
Ideas: 
- https://flutterigniter.com/env-specific-configuration/
- https://pub.dev/packages/environment_config
- https://www.dhiwise.com/post/flutter-envied-guide-to-managing-environment-variables

DigitalSeneca:
fixes #41 

## Description of changes
- To use different environment variables, dart requires the app to have different entry points (ie different files, main_prod.dart and main_dev.dart)
- This makes easier the future implementation of [Automate release of new versions to google play / apple store](#38) 

## How to use
- To start in prod: Run the file main_prod.dart, or in cmd: `flutter run -t lib/main_prod.dart`
- By default the app is started in dev mode. It can also be started in dev mode by running any of the following:
  - Run main.dart or main_dev.dart
  - Run `flutter run` or `flutter run -t lib/main_dev.dart`

### DIFF
--- a/assets/config/dev.json
+++ b/assets/config/dev.json
@@ -1,0 +1,3 @@
+{
+    "baseUrl": "https://madev.creaf.cat"
+}
--- a/assets/config/dev.json
+++ b/assets/config/dev.json
@@ -1,3 +1,3 @@
 {
-    "baseUrl": "https://madev.creaf.cat"
+    "baseUrl": "https://webdev.mosquitoalert.com"
 }
--- a/assets/config/prod.json
+++ b/assets/config/prod.json
@@ -1,0 +1,3 @@
+{
+    "baseUrl": "https://webserver.mosquitoalert.com"
+}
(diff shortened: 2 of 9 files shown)


assets/config/dev.json
```diff
@@ -0,0 +1,3 @@
+{
+    "baseUrl": "https://madev.creaf.cat"
```
epou, line 2: This is no longer the URL for the dev server. Change it to: `https://webdev.mosquitoalert.com`

DigitalSeneca: You're right, I pulled the branch from main instead of v33 so I mixed it by accident

epou (commented):
I'd rather go for using prod as default instead of dev, just to avoid accidentally releasing new apps with dev configurations.


By the way, could you take a look at https://github.com/flutter/flutter/pull/128668 ? It seems from flutter 3.13 there is native support for .env files (see `--dart-define-from-file`)

epou (approved):

PR #82, opened 2024-02-08
40 change endpoint fetching reports

### LINKED ISSUE #40: Change endpoint used for fetching reports
Use the API endpoint: `/api/reports?user=<USER_UUID_HERE>` instead of `/api/nearby_reports_nod`

Let the @Mosquito-Alert/web team know when about to apply the changes. They need to make sure the endpoint does not expose deleted reports.

DigitalSeneca:
fixes #40

### DIFF
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -36,7 +36,6 @@ class ApiSingleton {
 
   //Reports
   static const reports = '/reports/';
-  static const nearbyReports = '/nearby_reports_nod/';
 
   //Session
   static const sessions = '/sessions';
@@ -436,10 +435,9 @@ class ApiSingleton {
     try {
       var userUUID = await UserManager.getUUID();
 
-      final response = await http
-          .get(
+      final response = await http.get(
         Uri.parse(
-            '$serverUrl$nearbyReports?lat=$lat&lon=$lon&page=${page ?? '1'}&user=$userUUID&page_size=75&radius=${100}' +
+            '$serverUrl$reports?user=$userUUID' +
                 (show_hidden == true ? '&show_hidden=1' : '') +
                 (show_verions == true ? '&show_versions=1' : '')),
         headers: headers,
@@ -457,28 +455,13 @@ class ApiSingleton {
             'Request: ${response.request.toString()} -> Response: ${response.body}');
         return null;
       } else {
-        Map<String, dynamic> jsonAnswer = json.decode(response.body);
-        List<Report> list = [];
-        page = jsonAnswer['next'];
-
-        if (allReports == null) {
-          allReports = [];
-          UserManager.profileUUIDs = jsonAnswer['user_uuids'];
+        List<dynamic> jsonAnswer = json.decode(response.body);
+        var list = <Report>[];
+        for (var item in jsonAnswer) {
+          list.add(Report.fromJson(item));          
         }
-        for (var item in jsonAnswer['results']) {
-          if (UserManager.profileUUIDs.any((id) => id == item['user'])) {
-            list.add(Report.fromJson(item));
-          }
-        }
-        allReports.addAll(list);
-        //callback(list);
-        if (page == null) {
-          return allReports;
-        }
+        return list;
       }
-
-      return getReportsList(lat, lon,
-          page: page, allReports: allReports, radius: radius);
     } catch (e) {
       // print(e);
       return null;
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -17,14 +17,14 @@ import 'package:mosquito_alert_app/models/topic.dart';
 import 'package:mosquito_alert_app/utils/PushNotificationsManager.dart';
 import 'package:mosquito_alert_app/utils/UserManager.dart';
 import 'package:mosquito_alert_app/utils/Utils.dart';
+import 'package:mosquito_alert_app/app_config.dart';
 import 'package:path_provider/path_provider.dart';
 
 class ApiSingleton {
   static final _timeoutTimerInSeconds = 10;
 
-  static String devBASEURL = 'https://webdev.mosquitoalert.com';
-  static String baseUrl = 'https://webserver.mosquitoalert.com';
-  static String serverUrl = '$baseUrl/api';
+  static String baseUrl = '';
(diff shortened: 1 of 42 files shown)


lib/pages/my_reports_pages/my_reports_page.dart
```diff
@@ -579,206 +574,204 @@ class _MyReportsPageState extends State<MyReportsPage> {
                           ),
                         ],
                       ),
-                      isMine
```
DigitalSeneca, line 582: in the lines below I didn't change logic, just remove "isMine" which makes it seem like many lines were changed

lib/pages/my_reports_pages/my_reports_page.dart
```diff
@@ -579,206 +574,204 @@ class _MyReportsPageState extends State<MyReportsPage> {
...
+                                      fontSize: 14,
+                                      textAlign: TextAlign.center),
+                                ],
+                              ),
+                              SizedBox(
+                                height: 12,
+                              ),
+                            ],
+                          ),
+                          report.photos != null && report.photos!.isNotEmpty
```
DigitalSeneca, line 606: Here there's a change. Diff:
```
- report.photos!.isNotEmpty
+ report.photos != null && report.photos!.isNotEmpty
```

lib/api/api.dart
```diff
@@ -457,28 +445,13 @@ class ApiSingleton {
...
-          if (UserManager.profileUUIDs.any((id) => id == item['user'])) {
-            list.add(Report.fromJson(item));
-          }
-        }
-        allReports.addAll(list);
-        //callback(list);
-        if (page == null) {
-          return allReports;
+        List<dynamic> jsonAnswer = json.decode(response.body);
+        var list = <Report>[];
```
epou, line 449: `var List<Report> list = [];` ?

lib/pages/my_reports_pages/my_reports_page.dart
epou: Is the only change done here? In the pull request the diff are being displayed wrong

lib/pages/my_reports_pages/my_reports_page.dart
```diff
@@ -789,43 +782,47 @@ class _MyReportsPageState extends State<MyReportsPage> {
...
 
-      List<Report>? list = await ApiSingleton().getReportsList(
-          location!.latitude, location!.longitude,
-          radius: (distance ?? 1000 / 2).round());
+      var myData = await ApiSingleton().getReportsList();
 
-      if (list == null) {
-        list = [];
-      }
+      myData ??= [];
```
epou, line 792: Why this? haven't you already set myData to the output of the API report fetch?

lib/pages/my_reports_pages/my_reports_page.dart
```diff
@@ -836,7 +833,7 @@ class _MyReportsPageState extends State<MyReportsPage> {
                 myData[0].current_location_lon!)
             : LatLng(myData[0].selected_location_lat!,
                 myData[0].selected_location_lon!);
-        mapController.animateCamera(CameraUpdate.newCameraPosition(
+        await mapController.animateCamera(CameraUpdate.newCameraPosition(
             CameraPosition(target: location, zoom: 15)));
       }
```
epou, line 838: Looks like duplicated code after your changes...

lib/pages/my_reports_pages/my_reports_page.dart
DigitalSeneca: that's the same as 
```
if (list == null) {
        list = [];
}
```
but the dart linter showed me the following: 
- Prefer using `??=` over testing for null.dart[prefer_conditional_assignment](https://dart-lang.github.io/linter/lints/prefer_conditional_assignment.html)

DigitalSeneca: It can be null if the api response is not status code 200

lib/pages/my_reports_pages/my_reports_page.dart
DigitalSeneca: in this specific code snippet, yes

lib/pages/my_reports_pages/my_reports_page.dart
epou: Mmm... I was thinking of returning an empty list instead of null if the status code is not 200... In the end, you're doing the same.

Or make it one liner like: `var myData = await ApiSingleton().getReportsList() ?? [];`

But it's strange having a method that can return null if that null value is of no interest and casting again to an empty list when needed to be used.

lib/api/api.dart
DigitalSeneca: This would get this error
```
Variables can't be declared using both 'var' and a type name.
Try removing 'var.'dart(var_and_type)
```

This would work: `List<Report> list = [];`
but then we get: 
```
Omit type annotations for local variables.dart
[omit_local_variable_types](https://dart-lang.github.io/linter/lints/omit_local_variable_types.html)
```

epou: You're right, i wasn't aware of that!

lib/pages/my_reports_pages/my_reports_page.dart
```diff
@@ -579,206 +574,204 @@ class _MyReportsPageState extends State<MyReportsPage> {
...
+                                                              .question!
+                                                              .startsWith(
+                                                                  'question')
+                                                          ? MyLocalizations.of(
+                                                              context,
+                                                              report
                                                                   .responses![
                                                                       index]!
-                                                                  .answer
+                                                                  .answer)
```
epou, line 688: `.question)` here, no?

lib/pages/my_reports_pages/my_reports_page.dart
DigitalSeneca: yes, you're right. I'll refactor it

lib/pages/my_reports_pages/my_reports_page.dart
DigitalSeneca: yeah I got confused by the diff, I have reverted the commit

epou (commented):
Looks good to me :)

PR #83, opened 2024-02-08
Refactor MyReportsPage for more readability

DigitalSeneca:
Continuation of #82 to keep PRs smaller

Negate the if statement to have the small code block first, and after the bigger one with one less indentation.
In my opinion, that's more readable. No functionality was changed


epou (approved):
Agree

PR #84, opened 2024-02-09
phone_upload_time fields should be set on POST to API

### LINKED ISSUE #39: phone_upload_time fields should be set on POST to API
Report.phone_upload_time (only set when version_number == 0, otherwise, keep value)
 - To take into account that for version_number != 0, the fields report.phone_upload_time and report.version_time must be kept)
Fix.phone_upload_time

DigitalSeneca:
fixes #39 

## Description of changes
- phone_upload_time only set on createReport of the api
- On createReport, set phone_upload_time to now() if version_number == 0, otherwise keep value (if not null)

### DIFF
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -296,7 +296,7 @@ class ApiSingleton {
         body.addAll({'version_number': report.version_number});
         if (report.version_number == 0){
           body.addAll({'phone_upload_time': DateTime.now().toUtc().toIso8601String()});
-        }else if (report.phone_upload_time != null && report.phone_upload_time!.isNotEmpty){
+        } else if (report.phone_upload_time != null && report.phone_upload_time!.isNotEmpty){
           body.addAll({'phone_upload_time': report.phone_upload_time});
         }
       }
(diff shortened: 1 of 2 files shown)


lib/api/api.dart
```diff
@@ -294,17 +294,18 @@ class ApiSingleton {
       }
       if (report.version_number != null) {
         body.addAll({'version_number': report.version_number});
+        if (report.version_number == 0){
+          body.addAll({'phone_upload_time': DateTime.now().toUtc().toIso8601String()});
+        } else if (report.phone_upload_time != null && report.phone_upload_time!.isNotEmpty){
+          body.addAll({'phone_upload_time': report.phone_upload_time});
+        }
```
epou, line 301: `phone_upload_time` is required in the POST, i'm afraid if version_number is null will never set the field.

I would go to an independent if statement checking that version_number is set and is 0, then set phone_upload_time to now, keep same logic than before

DigitalSeneca: I've modified it. But there could be a case where version number is not 0 and phone_upload_time is empty, so it would be passed in the POST and the api call would fail?

epou: You're right... But that was the current flow since today... We'll need to improve this in the future.

DigitalSeneca: okay, so then we leave it like it's now for the current PR?

epou: Yep

PR #85, opened 2024-02-13
Add geocoding to display city in reports

### LINKED ISSUE #77: Use 'geocoding' package for displaying geo information
Displaying geo information was removed on the pull request: https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/34

Try to keep displaying that information by using the `placemarkFromCoordinates` from the `geocoding` library. (https://pub.dev/packages/geocoding) package instead.

Things that should display geo information again (that was removed on the pull request):

---

1. Show geo information on report list page.

https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/6c5993a230a86f958c8dca8bcfef2994a6b93ebe/lib/pages/my_reports_pages/my_reports_page.dart#L558

Should append: `"{CITY} ({SUB_ADMIN_AREA})"`

---

2. model `Report.displayCity` should be filled with the CITY name asynchronously.

It's used here: https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/6c5993a230a86f958c8dca8bcfef2994a6b93ebe/lib/pages/my_reports_pages/components/reports_list_widget.dart#L68

DigitalSeneca:
fixes #77 

| 1 | 2 |
| image  | image  |

### DIFF
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -795,7 +795,7 @@ class ApiSingleton {
   }
   
   Future<String> getCityNameFromCoords(double lat, double lon) async {
-    var locale = "${await UserManager.getLanguage()}_${await UserManager.getLanguageCountry()}";
+    var locale = '${await UserManager.getLanguage()}_${await UserManager.getLanguageCountry()}';
     var placemarks = await placemarkFromCoordinates(lat, lon, localeIdentifier: locale);
     return placemarks[0].locality ?? '';
   }
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -795,7 +795,8 @@ class ApiSingleton {
   }
   
   Future<String> getCityNameFromCoords(double lat, double lon) async {
-    var placemarks = await placemarkFromCoordinates(lat, lon);
+    var locale = "${await UserManager.getLanguage()}_${await UserManager.getLanguageCountry()}";
+    var placemarks = await placemarkFromCoordinates(lat, lon, localeIdentifier: locale);
     return placemarks[0].locality ?? '';
   }
 }
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -449,16 +449,6 @@ class ApiSingleton {
         List<dynamic> jsonAnswer = json.decode(response.body);
         var list = <Report>[];
         for (var item in jsonAnswer) {
-          double lat, lon;
-          if (item['location_choice'] == 'current') {
-              lat = item['current_location_lat'] ?? Utils.defaultLocation.latitude;
-              lon = item['current_location_lon'] ?? Utils.defaultLocation.longitude;
-          } else {
-              lat = item['selected_location_lat'] ?? Utils.defaultLocation.latitude;
-              lon = item['selected_location_lon'] ?? Utils.defaultLocation.longitude;
-          }
-          var cityName = await Utils.getCityNameFromCoords(lat, lon);
-          item['display_city'] = cityName;
           list.add(Report.fromJson(item));          
         }
         return list;
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -454,11 +454,11 @@ class ApiSingleton {
         for (var item in jsonAnswer) {
           double lat, lon;
           if (item['location_choice'] == 'current') {
-              lat = item['current_location_lat'];
-              lon = item['current_location_lon'];
+              lat = item['current_location_lat'] ?? Utils.defaultLocation.latitude;
+              lon = item['current_location_lon'] ?? Utils.defaultLocation.longitude;
           } else {
-              lat = item['selected_location_lat'];
-              lon = item['selected_location_lon'];
+              lat = item['selected_location_lat'] ?? Utils.defaultLocation.latitude;
+              lon = item['selected_location_lon'] ?? Utils.defaultLocation.longitude;
           }
           var cityName = await getCityNameFromCoords(lat, lon);
           item['display_city'] = cityName;
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -796,6 +796,6 @@ class ApiSingleton {
   
   Future<String> getCityNameFromCoords(double lat, double lon) async {
     var placemarks = await placemarkFromCoordinates(lat, lon);
-    return placemarks[0].locality ?? 'Error';
+    return placemarks[0].locality ?? '';
   }
 }
--- a/lib/api/api.dart
+++ b/lib/api/api.dart
@@ -4,6 +4,7 @@ import 'dart:io';
 
 import 'package:dio/dio.dart';
 import 'package:firebase_auth/firebase_auth.dart';
+import 'package:geocoding/geocoding.dart';
 
 import 'package:http/http.dart' as http;
 import 'package:http_parser/http_parser.dart';
@@ -330,6 +331,9 @@ class ApiSingleton {
       if (report.selected_location_lon != null) {
         body.addAll({'selected_location_lon': report.selected_location_lon});
       }
+      if (report.displayCity != null) {
+        body.addAll({'display_city': report.displayCity});
+      }
       if (report.package_name != null) {
         body.addAll({'package_name': report.package_name});
       }
@@ -448,6 +452,16 @@ class ApiSingleton {
         List<dynamic> jsonAnswer = json.decode(response.body);
         var list = <Report>[];
         for (var item in jsonAnswer) {
+          double lat, lon;
+          if (item['location_choice'] == 'current') {
+              lat = item['current_location_lat'];
+              lon = item['current_location_lon'];
+          } else {
+              lat = item['selected_location_lat'];
+              lon = item['selected_location_lon'];
+          }
+          var cityName = await getCityNameFromCoords(lat, lon);
+          item['display_city'] = cityName;
           list.add(Report.fromJson(item));          
         }
         return list;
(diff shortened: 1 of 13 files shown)


lib/pages/forms_pages/components/biting_location_form.dart
```diff
@@ -47,7 +47,8 @@ class _BitingLocationFormState extends State<BitingLocationForm> {
                   Utils.report!.selected_location_lon!)));
           currentLocation = Position(
               latitude: Utils.report!.selected_location_lat!,
-              longitude: Utils.report!.selected_location_lon!);
+              longitude: Utils.report!.selected_location_lon!,
+              accuracy: -1, altitude: -1, altitudeAccuracy: -1, speed: -1, speedAccuracy: -1, heading: -1, headingAccuracy: -1, timestamp: null);
```
DigitalSeneca, line 51: I got the error that all parameters are required and I could not compile otherwise. I am not sure why this was not happening before, probably some version/dependencies stuff
```
The named parameter 'speed' is required, but there's no corresponding argument.
Try adding the required argument.dart[missing_required_argument](https://dart.dev/diagnostics/missing_required_argument)
```
image

lib/utils/Utils.dart
```diff
@@ -104,7 +104,7 @@ class Utils {
       }
     }
 
-    if (session!.id != null && language != null) {
```
DigitalSeneca, line 107: dart linter: "language can't be null, remove useless comparison"

lib/api/api.dart
```diff
@@ -779,4 +793,10 @@ class ApiSingleton {
       return false;
     }
   }
+  
+  Future<String> getCityNameFromCoords(double lat, double lon) async {
+    var locale = '${await UserManager.getLanguage()}_${await UserManager.getLanguageCountry()}';
```
DigitalSeneca, line 798: Locale in geocoding is optional, if not provided it uses the one that the phone has selected
If we add it, it has to be in the formal language_country (ie es_ES, de_DE, en_US...)

lib/api/api.dart
```diff
@@ -779,4 +793,10 @@ class ApiSingleton {
...
+    var placemarks = await placemarkFromCoordinates(lat, lon, localeIdentifier: locale);
+    return placemarks[0].locality ?? '';
```
DigitalSeneca, line 800: If more than one placemark is returned, all of them will be very close to each other, so I'd just take the first one

lib/api/api.dart
```diff
@@ -331,6 +332,9 @@ class ApiSingleton {
       if (report.selected_location_lon != null) {
         body.addAll({'selected_location_lon': report.selected_location_lon});
       }
+      if (report.displayCity != null) {
+        body.addAll({'display_city': report.displayCity});
+      }
```
epou, line 337: The API does not accept this parameter... Need to be removed. In the future we will make the backend fetch the city name and expose it through the API.

lib/api/api.dart
```diff
@@ -449,6 +453,16 @@ class ApiSingleton {
...
+          double lat, lon;
+          if (item['location_choice'] == 'current') {
+              lat = item['current_location_lat'] ?? Utils.defaultLocation.latitude;
+              lon = item['current_location_lon'] ?? Utils.defaultLocation.longitude;
+          } else {
+              lat = item['selected_location_lat'] ?? Utils.defaultLocation.latitude;
+              lon = item['selected_location_lon'] ?? Utils.defaultLocation.longitude;
+          }
+          var cityName = await getCityNameFromCoords(lat, lon);
+          item['display_city'] = cityName;
```
epou, line 465: I suggest relocating this code within the Report constructor as it seems more fitting there. This logic appears unrelated to the communication with the API.

lib/api/api.dart
```diff
@@ -779,4 +793,10 @@ class ApiSingleton {
       return false;
     }
   }
+  
+  Future<String> getCityNameFromCoords(double lat, double lon) async {
```
epou, line 797: I think it'd be better to place this function in the Utils folder since it's note related to how to communicate with the API.

lib/api/api.dart
epou: Would you mind creating a function in the UserManger called getUserLocale (or something like this) that returns this value? It could be worth reusing that function in the future for other cases instead of contacting the result of getLanguage and getLanguageCountry always

epou: It would be nice to check if it works in basque... I was wondering if eu_ES is the correct locale code. If it's prone to error, only using language is also accepted in the function `placemarkFromCoordinates`

lib/api/api.dart
epou: Need to check if placemarks is empty before getting the first position:
```dart
if (placemarks.isNotEmpty) {
    // code here
 }
 ```

File: lib/pages/forms_pages/components/biting_location_form.dart (hunk 1 above)
Comment (epou) [reply lib/pages/forms_pages/components/biting_location_form.dart:51]: Ah, that's interesting. It appears that the `Position` class from the `geolocator` package wasn't designed to be instantiated directly. It serves as the type of object returned by methods like `getCurrentPosition` or `getLastKnownPosition`. Considering this, it might be more appropriate to replace them all and utilize the `LatLng` class from `google_maps_flutter.dart` instead, which seems to be used elsewhere in the project. However, making this change is significant and requires thorough testing on the device to ensure that everything continues to function as intended.

File: lib/api/api.dart (hunk 4 above)
Comment (DigitalSeneca) [reply lib/api/api.dart:800]: "locality" can be null, for example if it's in the middle of the ocean (0, 0). In this case I assume in the app it will just not show anything. I can modify the method signature to be able to return null values:
`Future<String?> getCityNameFromCoords(double lat, double lon) async {`

Another option is to do something like this to ensure that always either the locality or an empty string is returned
```
Future<String> getCityNameFromCoords(double lat, double lon) async {
    var locale = '${await UserManager.getLanguage()}_${await UserManager.getLanguageCountry()}';
    var placemarks = await placemarkFromCoordinates(lat, lon, localeIdentifier: locale);
    if (placemarks.isEmpty) { return ''; }
    return placemarks.first.locality ?? '';
}
```

File: lib/api/api.dart (hunk 6 above)
Comment (DigitalSeneca) [reply lib/api/api.dart:465]: I agree, I tried first to do it first in the Report.fromJson constructor but got:
```
The modifier 'async' can't be applied to the body of a constructor.
Try removing the modifier.dart[invalid_modifier_on_constructor](https://dart.dev/diagnostics/invalid_modifier_on_constructor)
```

File: lib/api/api.dart (hunk 4 above)
Comment (epou) [reply lib/api/api.dart]: Yeah, returning always String looks good to me

File: lib/api/api.dart (hunk 3 above)
Comment (DigitalSeneca) [reply lib/api/api.dart]: > Would you mind creating a function in the UserManger called getUserLocale (or something like this) that returns this value?

done

> It would be nice to check if it works in basque... I was wondering if eu_ES is the correct locale code. If it's prone to error, only using language is also accepted in the function placemarkFromCoordinates

Seems like it's not working as expected. It's returning the name in the countries name instead of in selected locale (Mountain View in US; München in Germany, and Bilbo in Basque), when I have basque as my language in the app
image

Comment (DigitalSeneca) [reply lib/api/api.dart]: > Or try using official method: https://api.flutter.dev/flutter/intl/Intl/getCurrentLocale.html

```
static String getCurrentLocale() {
  return defaultLocale ??= systemLocale;
}
```

But then with this method, if the user selects another language within the app, it will always still show the systemLocale

File: lib/api/api.dart (hunk 6 above)
Comment (epou) [reply lib/api/api.dart]: Have you tried using a separate async function inside the class and call it?

```dart
class Report {
 // properties here

  Report() {
   // current code here
    _initialize();
  }

  Future<void> _initialize() async {
   // code to get appropiate lat and lon here
    var cityName = await getCityNameFromCoords(lat, lon);
    displayCity = cityName;
  }
}
```

lib/api/api.dart
epou: Ah yes, you're right! The only thing I'm afraid of is the `UserManger.getUserLocale` returning a combination of language + country that is not a valid locale and the code fails. Maybe it's safer to just use the language code as stated in the [docu](https://pub.dev/packages/geocoding) ('en' for English)

DigitalSeneca: > The localeIdentifier should be formatted using the syntax: [languageCode]_[countryCode]

`en` seems to be the only option where a country code is not needed. If we are afraid, maybe the safest option is to leave this option parameter empty, so it will use the system's language

epou: As far as this codes are always valid, it's gonna be ok
https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/6c5993a230a86f958c8dca8bcfef2994a6b93ebe/lib/pages/settings_pages/settings_page.dart#L38-L62

lib/api/api.dart
DigitalSeneca: I just tried it and it works, nice. I agree that's better than it being in the api.dart

lib/pages/forms_pages/components/biting_location_form.dart
DigitalSeneca: yup, LatLng seems more appropriate when the other parameters are not used

> However, making this change is significant and requires thorough testing on the device to ensure that everything continues to function as intended.

Are we doing it in a future version then?

epou: Could you try replacing all "Position" instances for LatLng? Maybe it's just that simple. Test it in your emulator just to be sure, but we can test it later once the draft is released

DigitalSeneca: I just tried it and seems to be working on my emulator. You can see it in the latest commit "Replace Position for LatLng"

epou (approved):

PR #86, opened 2024-02-15
Update translations

epou:


update_locales.py
DigitalSeneca, line 1: Nice, this makes it faster/convenient
Is your idea to later remove it with https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/issues/69 or to keep it so there's always this option?

lib/pages/my_reports_pages/my_reports_page.dart
```diff
@@ -486,7 +486,7 @@ class _MyReportsPageState extends State<MyReportsPage> {
                                             children: [
                                               Style.body(
                                                   MyLocalizations.of(
-                                                      context, 'more_info'),
+                                                      context, 'more_info') + ': ',
```
DigitalSeneca, line 489: I can't compile the project
```
The operator '+' can't be unconditionally invoked because the receiver can be 'null'.
Try adding a null check to the target ('!').dart[unchecked_use_of_nullable_value](https://dart.dev/diagnostics/unchecked_use_of_nullable_value)
String +(String other)
```

This fixes it, it has to be added in other places in this file too
```
context, 'more_info')! + ': ',
```

assets/language/en_US.json
DigitalSeneca, line 1: If we move from en.json to en_US.json we also need to adapt the code in other places like `class MyLocalizations` and `MyLocalizationsDelegate`
  ```
static Future<void> _loadEnglishTranslations() async {
    var jsonContent = await rootBundle.loadString('assets/language/en.json');
    _englishValues = json.decode(jsonContent);
  }
```
```
Exception has occurred.
FlutterError (Unable to load asset: assets/language/en.json)
```

**Review by DigitalSeneca — CHANGES_REQUESTED** [review]

File: update_locales.py
Comment (epou) [reply update_locales.py:1]: My idea was to use both... one that run automatically, the other can be run manually

File: lib/pages/my_reports_pages/my_reports_page.dart (hunk 1 above)
Comment (epou) [reply lib/pages/my_reports_pages/my_reports_page.dart]: Fixed on [bb85100](https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/86/commits/bb8510037f75befda5a2ece3a1d7ffeae831fdb6)

File: assets/language/en_US.json
Comment (epou) [reply assets/language/en_US.json:1]: Fixed on [e6bf563](https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/86/commits/e6bf563eb25dd3fc589499b8ddd3b78a2d51a17b)

File: assets/language/de_DE.json [hunk 2]
```diff
@@ -304,23 +304,26 @@
...
-    "disable_auto_hashtag_text": "Auto-hashtag will be turned off. Are you sure?",
-    "enable_auto_hashtag_text": "This is an automatic way of adding a specific hashtag to all your records. Use it only if you want to always have your records identified  by a unique hashtag.",
-    "enable_auto_hashtag_hint": "Write your hashtag.\nRemember not to include personal information.",
-    "save": "Save",
-    "ma_not_available": "Looks like Mosquito Alert isn't available right now. Please, try again later.",
-    "not_enough_permissions_location": "We don't have enough permissions to acess the location data on your device. You can enable this on the app settings. Want to go there?",
+    "disable_auto_hashtag": "",
+    "enable_auto_hashtag": "",
+    "disable_auto_hashtag_text": "",
+    "enable_auto_hashtag_text": "",
```
DigitalSeneca, line 317: The translate() method defaults to english when the string is null, but in this case is empty so it just shows it
https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/de67a3d10ee7afb955923592ed74da1fbaea6cd1/lib/utils/MyLocalizations.dart#L30-L32
image
image

So if we're going to have the empty strings in the json, we have to adapt it
```
String translate(String? key) { 
    if (_localisedValues[key] != null && _localisedValues[key].isNotEmpty) { return _localisedValues[key] }
   return _englishValues[key] ?? ''; 
} 
```

DigitalSeneca (commented):
Except from the other comment, seems fine to me

assets/language/de_DE.json
epou: Fixed in [d5c7229](https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/86/commits/d5c722996314fdc70adf5ac97c02290e9332a9f2)

DigitalSeneca (approved):

PR #87, opened 2024-02-16
Small bugs

epou:


lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -14,6 +14,10 @@ class ReportsList extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+
+    // Sort the reports by creation_time in descending order
+    reports.sort((a, b) => b.creation_time.compareTo(a.creation_time));
```
DigitalSeneca, line 19: creation_time can be null so I can't compile

epou: Can I make it non-nullable? The DB is not allowed null for this field, so theoretically it should never be null

DigitalSeneca: yes, we can try it

epou: Fixed in [d900c1d](https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/87/commits/d900c1da902cdc267e354cfbde98ea8d855c065e)

DigitalSeneca (approved):

PR #94, opened 2024-02-21
Call forceUpdateMap to avoid calling updateAggregatedPoints

epou:


DigitalSeneca (approved):
seems to be working fine for me

PR #97, opened 2024-03-08
Background tracking takes 5 snapshots per day

### LINKED ISSUE #9: Check optional background tracking and make sure exactly 5 tracks per day are being taken
The tracking currently relies on Flutter Background Geolocation (https://pub.dev/packages/flutter_background_geolocation), but enforcing the 5-sample per day requirement was complicated when first implemented.

DigitalSeneca:
fixes #9

## Description of changes
- Remove flutter_background_geolocation in favor of geolocator
- Send 5 snapshots per day (one every 4h 48min)
- Permission handling: 
  - On the one hand, standard permission handling by the system (must be set to always)
  - On the other hand, user can enable/disable tracking in the settings
  - Permission must be enabled in both places for the bg tracking to get locations and send it to the server

image


lib/main.dart
```diff
@@ -22,9 +27,40 @@ void main({String env = 'prod'}) async {
...
 }
 
+void callbackDispatcher() {
+  Workmanager().executeTask((task, inputData) async {
+    await Firebase.initializeApp();
+
+    switch (task) {
+      case 'backgroundTracking':
+        var permission = await Geolocator.checkPermission();
+        var isBgTrackingEnabled = await UserManager.getTracking();
```
epou, line 48: Could you double check that `UserManager.getTracking()` return True when tracking is enabled? I'm getting confused in the `setTracking` method, which seems to set the variable `trackingDisabled`... Not sure

epou: Ups... nevermind, I hadn't seen that you modified the UserManager when i posted the comment...

lib/pages/settings_pages/settings_page.dart
```diff
@@ -146,10 +147,9 @@ class _SettingsPageState extends State<SettingsPage> {
               height: 10,
             ),
             SettingsMenuWidget(
-                enableTracking!
-                    ? MyLocalizations.of(context, 'enable_background_tracking')
-                    : MyLocalizations.of(
-                        context, 'disable_background_tracking'), () {
+                isBgTrackingEnabled
+                    ? MyLocalizations.of(context, 'disable_background_tracking')
+                    : MyLocalizations.of(context, 'enable_background_tracking'), () {
```
epou, line 152: How would you see replacing this 'dialog box' with a toggle switch, continuing with permissions management as it is done now?

DigitalSeneca: sure, I can take a look into adding a toggle switch

epou: Great! To make things easier, let's use the `Colors.orange` color when the switch is active. Also, I think a toggle switch would work better for boolean options. By the way, we could save some time by reusing the Material implementation of the switch.

epou:
We must ensure that we send masked lat/lon to our API.

Here:
https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/72ced5d1b88297fbbc4671f15908ab382789aa8e/lib/api/api.dart#L495-L496

Doing: 
```dart
'masked_lat' = (lat / Utils.maskCoordsValue).floor() * Utils.maskCoordsValue;
'masked_lon' = (lon / Utils.maskCoordsValue).floor() * Utils.maskCoordsValue;
```

lib/pages/settings_pages/settings_page.dart
```diff
@@ -145,13 +144,18 @@ class _SettingsPageState extends State<SettingsPage> {
...
             ),
-            SettingsMenuWidget(
-                enableTracking!
-                    ? MyLocalizations.of(context, 'enable_background_tracking')
-                    : MyLocalizations.of(
-                        context, 'disable_background_tracking'), () {
-              _disableBgTracking();
-            }),
+            SwitchListTile(
+              title: Text(MyLocalizations.of(context, 'enable_background_tracking') ?? ''),
```
epou, line 148: change 'enable_background_tracking' to 'background_tracking_title'

PR #98, opened 2024-03-20
Changes for IoS deployment

StuffJoy:
1. Ran Dart fix for minor issues
2. Made changes listed below to remove deprecated/broken code (the app would not debug and run without addressing them) 
- raisedbutton depreciated and causing issues, elevated button used instead 
- flatbutton not found in material import, textbutton used instead 
- few future methods needed a throw statement for error handling 
- scaffold.of won’t work because…reasons. I replaced it with ScaffoldMessenger.of

3. Updated minimum target device from 10.3 to 12 (anything below 11 and the app will not work)
4. Open the iOS/runner project.xcodeproj in Xcode and change target device from 10.3 to 12
5. Add GoogleService-Info.plist and info.plist files
6. Updated firebase auth from 4.3.0 to 4.3.1 (changes were made in the 4.3.1 version that resolves the issue experienced)

System changes
Upgraded ruby to 3.2.2 (anything under 2.6 version clashes with cocoa pods)
Updated cocoa pods

### DIFF
--- a/lib/utils/Utils.dart
+++ b/lib/utils/Utils.dart
@@ -25,7 +25,6 @@ import 'package:shared_preferences/shared_preferences.dart';
 import 'package:url_launcher/url_launcher.dart';
 import 'package:uuid/uuid.dart';
 
-import '../models/question.dart';
 import 'MyLocalizations.dart';
 
 class Utils {
@@ -57,9 +56,7 @@ class Utils {
   };
 
   static void saveImgPath(File img) {
-    if (imagePath == null) {
-      imagePath = [];
-    }
+    imagePath ??= [];
     imagePath!
         .add({'image': img.path, 'id': report!.version_UUID, 'imageFile': img});
   }
@@ -82,13 +79,13 @@ class Utils {
     if (session == null) {
       reportsList = [];
 
-      String? userUUID = await UserManager.getUUID();
+      var userUUID = await UserManager.getUUID();
 
       dynamic response = await ApiSingleton().getLastSession(userUUID);
       if (response is bool && !response) {
         print('Unable to get last session.');
         return false;
-      } else if(response is ApiResponse){
+      } else if (response is ApiResponse) {
         print('response is of type ApiResponse, not a number.');
         return false;
       } else {
@@ -117,7 +114,7 @@ class Utils {
           session: session!.id,
           responses: []);
 
-      PackageInfo packageInfo = await PackageInfo.fromPlatform();
+      var packageInfo = await PackageInfo.fromPlatform();
       report!.package_name = packageInfo.packageName;
       report!.package_version = 33;
 
@@ -228,13 +225,13 @@ class Utils {
       return;
     }
 
-    List<Question?>? _questions = report!.responses;
+    var _questions = report!.responses;
 
     // add total bites
 
     if (question_id == 1) {
-      int currentIndex = _questions!.indexWhere((question) =>
-          question!.question_id == question_id);
+      var currentIndex = _questions!
+          .indexWhere((question) => question!.question_id == question_id);
       if (currentIndex == -1) {
         _questions.add(Question(
           question: question.toString(),
@@ -250,8 +247,8 @@ class Utils {
 
     //increase answer_value question 2
     if (question_id == 2) {
-      int currentIndex = _questions!.indexWhere((question) =>
-          question!.answer_id == answer_id);
+      var currentIndex = _questions!
+          .indexWhere((question) => question!.answer_id == answer_id);
       if (currentIndex == -1) {
         _questions.add(Question(
           question: question.toString(),
@@ -261,7 +258,7 @@ class Utils {
           answer_value: '1',
         ));
       } else {
-        int value = int.parse(_questions[currentIndex]!.answer_value!);
+        var value = int.parse(_questions[currentIndex]!.answer_value!);
         value = value + 1;
         _questions[currentIndex]!.answer_value = value.toString();
       }
@@ -276,7 +273,7 @@ class Utils {
         if (_questions.any(
             (q) => q!.question_id == question_id && q.answer_id != answer_id)) {
           //modify question
-          int index =
+          var index =
               _questions.indexWhere((q) => q!.question_id == question_id);
           _questions[index]!.answer_id = answer_id;
           _questions[index]!.answer = answer;
@@ -298,7 +295,7 @@ class Utils {
   }
 
   static void resetBitingQuestion() {
-    List<Question?> _questions = report!.responses!;
+    var _questions = report!.responses!;
 
     _questions.removeWhere((q) => q!.question_id == 2);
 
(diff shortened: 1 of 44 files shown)


lib/utils/Utils.dart
```diff
@@ -1128,10 +1139,10 @@ class Utils {
     }
   }
 
-  static getLanguage() {
-    if (ui.window != null && ui.window.locale != null) {
-      String stringLanguange = ui.window.locale.languageCode;
-      String? stringCountry = ui.window.locale.countryCode;
+  static ui.Locale getLanguage() {
```
epou, line 1142: not sure why `ui.Locale` added here. @DigitalSeneca any opinion?

epou (commented):
Hey! Thanks @StuffJoy for your pull request!

Just a heads up, we'll need to double-check when testing on our phones:
- removeBite() in lib/pages/forms_pages/components/biting_form.dart
- user ID copying on the settings still working good across all devices at lib/pages/settings_pages/settings_page.dart (lines 124-142).

Before we hit that merge button, though, I want to make sure @DigitalSeneca  gives the Android build the green light.

Thanks again for your hard work! Let's keep pushing forward!

P.S: I've been thinking it might be a good idea for us to get on the same page with our linter situation. Maybe we could standardize it across the team? That way, we can avoid these kinds of pull requests where most of the changes are just linter-related, not to mask the core changes with the linter ones.

lib/utils/style.dart
```diff
@@ -103,19 +103,17 @@ class Style {
...
-          ? color.withOpacity(0.3)
-          : colorPrimary.withOpacity(0.3),
-      shape: RoundedRectangleBorder(
-        borderRadius: BorderRadius.circular(3),
+      style: ElevatedButton.styleFrom(
+        elevation: 0,
+        tapTargetSize: MaterialTapTargetSize.shrinkWrap,
+        padding: EdgeInsets.symmetric(vertical: 14),
+        backgroundColor: color ?? colorPrimary,
+        foregroundColor: textColor ?? Colors.white,
```
DigitalSeneca, line 113: I am getting the following error on multiple places of this file for both `backgroundColor` and `foregroundColor`. It's most likely due to a different version of flutter/dart or dependencies
```
The named parameter 'backgroundColor' isn't defined.
Try correcting the name to an existing named parameter's name, or defining a named parameter with the name 'backgroundColor'.dart[undefined_named_parameter](https://dart.dev/diagnostics/undefined_named_parameter)
```

lib/utils/Utils.dart
DigitalSeneca: It's actually returning a Locale, both of those options would be correct I think

```
static Locale getLanguage() {
static ui.Locale getLanguage() {
```

DigitalSeneca (commented):
Thanks for the contribution @StuffJoy !

> Before we hit that merge button, though, I want to make sure @DigitalSeneca gives the Android build the green light.

Seems like we're having some conflict here with firebase_auth, so I can't compile on Android after the dependencies upgrade. Are you using a different version of flutter/dart?
```
The current Dart SDK version is 2.17.0.

Because firebase_auth >=4.4.0 requires SDK version >=2.18.0 <4.0.0 and no versions of firebase_auth match >=4.3.1 <4.4.0, firebase_auth >=4.3.1 is forbidden.
So, because mosquito_alert_app depends on firebase_auth ^4.3.1, version solving failed.
Running "flutter pub get" in Mosquito-Alert-Mobile-App...               
pub get failed (1; So, because mosquito_alert_app depends on firebase_auth ^4.3.1, version solving failed.)
```

> P.S: I've been thinking it might be a good idea for us to get on the same page with our linter situation. Maybe we could standardize it across the team? That way, we can avoid these kinds of pull requests where most of the changes are just linter-related, not to mask the core changes with the linter ones.

Yes, also @StuffJoy if you make multiple smaller pull requests divided by logic, this makes the life of the reviewers easier, and it's also easier for us to both catch issues and provide useful feedback. For example:
- PR1: Lint changes
- PR2: Minimum changes needed to compile on iOS
- PR3: few future methods needed a throw statement for error handling

lib/utils/style.dart
StuffJoy: No problem @DigitalSeneca. I am now reviewing the feedback

> Seems like we're having some conflict here with firebase_auth, so I can't compile on Android after the dependencies upgrade. Are you using a different version of flutter/dart? 

Yes it looks like I am using a more recent version. See below.

flutter --version
Flutter 3.13.7 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 2f708eb839 (6 months ago) • 2023-10-09 09:58:08 -0500
Engine • revision a794cf2681
Tools • Dart 3.1.3 • DevTools 2.25.0

DigitalSeneca:
> Would it be possible/feasible for you to upgrade to a more recent version and try again?

Thanks for providing the exact version you're using. I have updated my version to 3.13.7 and tried to compile but it is still not working with the following error message:

```
FAILURE: Build failed with an exception.

* Where:
Build file 'C:\Users\DigitalSeneca\AppData\Local\Pub\Cache\hosted\pub.dev\file_picker-5.3.1\android\build.gradle' line: 24

* What went wrong:
A problem occurred evaluating project ':file_picker'.
> No signature of method: build_8yy0krq21hzp6uwdyqcb5tvuy.android() is applicable for argument types: (build_8yy0krq21hzp6uwdyqcb5tvuy$_run_closure2) values: [build_8yy0krq21hzp6uwdyqcb5tvuy$_run_closure2@158715a2]

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 5s
Running Gradle task 'assembleDebug'...                              6,1s
Exception: Gradle task assembleDebug failed with exit code 1
```

So it seems an issue related to the version of `file_picker` and/or `android/build.gradle` but I could not fix it on a first attempt.
I tried upgrading file_picker to a higher version, but `6.0.0` or higher is not possible without having to update many other dependencies too

> The initial code used "primary" to set the background color of the button, however my editor (Visual Studio) would not allow me to compile the app without changing the button set from "RaisedButton" to "ElevatedButton", which uses backgroundColor to define the appearance of the button. Same for "onPrimary" and "foregroundColor".

Yes, on version 3.0.0 they were marked as "will be deprecated in a future version" but we could still use them, and in 3.13.7 they were finally removed so you had to update them

> I apologise for this aspect. The extent of changes necessary was unexpected so I did not consider a structured approach when I began. To clarify, are you requesting that I resubmit the pull request in phases, or is this advice for handling future pull requests?

If we can compile it on Android too, then I'd say to apply it for future pull requests, but it's not working at the moment unfortunately :(

Was it not possible to prepare the iOS changes on flutter 3.0.0? In this version all dependencies on pubspec.yaml are compatible

DigitalSeneca:
Hey @StuffJoy ! I checked and according to this answer, "application packages" should have the pubspec.lock added to git https://stackoverflow.com/a/16136740

PR #115, opened 2024-04-18
Notification bell

### LINKED ISSUE #43: Add notification bell counter
Display a badge with the number of unread notifications in the top right of the notification bell.

The app logo have also a badge number with the unread notifications

DigitalSeneca:
fixes #43


lib/pages/main/main_vc.dart
```diff
@@ -79,6 +82,20 @@ class _MainVCState extends State<MainVC> {
     }
   }
 
+  void _getNotificationCount() async {
+    List<MyNotification> notifications = await ApiSingleton().getNotifications();
+    var unacknowledgedCount = notifications.where((notification) => notification.acknowledged == false).length;
+    setState(() {
+      unreadNotifications = unacknowledgedCount;
+    });
```
epou, line 90: What do you think about using `updateNotificationCount(unacknowledgedCount)` , just for the sake of avoiding duplicating code..

lib/pages/notification_pages/notifications_page.dart
```diff
@@ -27,15 +28,18 @@ class NotificationsPage extends StatefulWidget {
 class _NotificationsPageState extends State<NotificationsPage> {
   List<MyNotification> notifications = [];
   StreamController<bool> loadingStream = StreamController<bool>.broadcast();
+  int updatedUnreadNotificationsCount = 1;
```
epou, line 31: Is this variable used?

DigitalSeneca: I've cleaned it now

epou:
Hey! Does the counter badge "break" visually when the count reaches two digits?

DigitalSeneca:
> Hey! Does the counter badge "break" visually when the count reaches two digits?

Seems to be working fine visually with 2 digits. Screenshot:
image

epou:
Cool! Can we fetch the data from the API in the `main_vc.dart` asynchronously and pass the list into the NotificationsPage view? This API call is quite slow and would be nice to be called every X minutes in the background and on each app start to avoid waiting every time the user goes to the notification page.

DigitalSeneca:
> Cool! Can we fetch the data from the API in the `main_vc.dart` asynchronously and pass the list into the NotificationsPage view? This API call is quite slow and would be nice to be called every X minutes in the background and on each app start to avoid waiting every time the user goes to the notification page.

yes, I'll do that. The minimum amount of time between periodic background tasks is every 15min for Android, is that good or should I do a different interval?

epou:
I didn't know about the 15min constraint for background tasks... Maybe it's too much. In that case, you can keep it as is.

By the way, I've seen there is a native class for badges in material. It would be worth a try:
- https://api.flutter.dev/flutter/material/Badge-class.html
- https://m3.material.io/components/badges/overview

epou:
Looks good to me! 
I feel curious if this https://api.flutter.dev/flutter/material/Badge-class.html might help in making the code easier, I feel the native library might consider a broad number of case scenarios and less prone to bugs.

Just to be consistent with the linked issue (#43), I'm not sure if this is working or not:
> The app logo have also a badge number with the unread notifications

I've found this, which might be helpful: https://stackoverflow.com/questions/61979970/how-to-show-a-notification-badge-on-the-app-icon-using-flutter

DigitalSeneca:
I'm sorry I didn't address the comment about the badge before requesting a new review. I'll pay more attention next time before asking for a new review

Here's the screenshot after the changes. I've also increased the bell size for it to be similar to the adn image next to it

| with unread | without unread |
| image | image |

lib/pages/main/main_vc.dart
```diff
@@ -118,15 +134,20 @@ class _MainVCState extends State<MainVC> {
...
-                    Navigator.push(
-                      context,
-                      MaterialPageRoute(
-                          builder: (context) => NotificationsPage()),
-                    );
-                  },
+                badges.Badge(
+                  position: badges.BadgePosition.topEnd(top: 2, end: 2),
+                  showBadge: unreadNotifications > 0,
+                  badgeContent: Text('$unreadNotifications'),
```
epou, line 140: Changing text color to white
```
Text(
        '$unreadNotifications',
        style: TextStyle(color: Colors.white),
      )
```

lib/pages/main/main_vc.dart
```diff
@@ -118,15 +134,20 @@ class _MainVCState extends State<MainVC> {
...
+                  child: IconButton(
+                    padding: EdgeInsets.only(top: 6),
+                    icon: Icon(Icons.notifications, size: 32, ),
```
epou, line 143: Could you try with a smaller size? I see the badge number too large

lib/pages/main/main_vc.dart
DigitalSeneca: I've changed it to white

epou (approved):

PR #117, opened 2024-04-22
Added fallback=auto support

epou:


update_locales.py
```diff
@@ -19,14 +30,21 @@ def update_locales(api_key: str, path: Path, format: str) -> None:
...
+        if is_poorly_translated and locale_code not in FORCE_LANGUAGES:
+            continue
+
         file_name = f"{locale_code}.{format}"
         file_path = path / file_name
 
-        response = session.get(f'https://localise.biz/api/export/locale/{file_name}')
+        params = {}
+        if locale_code in AUTO_FALLBACK:
+            params['fallback'] = 'auto'
```
DigitalSeneca, line 45: Do I understand correctly that the purpose of the fallback is that those languages `['es'_UY, 'eu_ES', 'ca_ES', 'gl_ES']` default to `es_ES` if there is no string provided? I am assuming the default `es_ES` is defined within localise.biz

DigitalSeneca (approved):
nice! I've tested it and seems to be working as expected for me. I've created #118 based on this

update_locales.py
epou: Yes! You're right, it's exactly like you say. The fallback=auto behaviour is configured in localise.biz

PR #118, opened 2024-04-22
Add Spanish (Uruguay) and update translation strings

DigitalSeneca:


epou:
Nice! but `background_tracking_title` and `background_tracking_subtitle` are from v34 branch and should not be copied in this pull request. I've added a new string for `es_UY` which is not in the pull request, so the script should be rerun to take them into account. We can make it once Urugauy translators confirm us they have finished.

DigitalSeneca:
> Nice! but `background_tracking_title` and `background_tracking_subtitle` are from v34 branch and should not be copied in this pull request. I've added a new string for `es_UY` which is not in the pull request, so the script should be rerun to take them into account. We can make it once Urugauy translators confirm us they have finished.

When I checkout `main` branch and run the script, that's what I get. Do I have to pass a different parameter to the command? I am not aware if localise.biz also uses branches internally

As a side note, if those 2 v34 strings are present in v33, it should not cause issues in functionality nor UI even if it's not technically correct

The string es_UY appears in the PR, but it's at the bottom of the json files

epou (changes requested):
Still need to modify Spanish -> Spanish (Spain) + add Spanish (Uruguay) in the list. See below:

https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/b548cdc6a176491b6a314eef0fede9bb672a28d4/lib/pages/settings_pages/settings_page.dart#L38C9-L62

lib/pages/settings_pages/settings_page.dart
```diff
@@ -42,7 +42,8 @@ class _SettingsPageState extends State<SettingsPage> {
     Language('de_DE', 'German'),
     Language('el_GR', 'Greek'),
     Language('en_US', 'English'),
-    Language('es_ES', 'Spanish'),
+    Language('es_ES', 'Spanish (Spain)'),
```
epou, line 45: This should be 'Spanish' only, without Spain

DigitalSeneca: yes, I just pushed the commit with the fixed strings!

DigitalSeneca: I missed it, it's without "(Spain)" now!

epou (approved):

PR #119, opened 2024-04-23
Open selected notification

### LINKED ISSUE #59: Make sure that for push notifications, tapping on the notification opens the related message in the app
We should check that the push notification from the server sends the appropiate payload so that the app can catch and open the appropiate notification id inside the app.

DigitalSeneca:
fixes #59

push-notification-open-correctly


epou (approved):
Bravo! Does it work in iOS?

DigitalSeneca:
Closed in favor of #121 to merge it to main instead of v34

PR #121, opened 2024-04-23
Open selected notification

### LINKED ISSUE #59: Make sure that for push notifications, tapping on the notification opens the related message in the app

DigitalSeneca:
fixes #59

push-notification-open-correctly


epou (commented):
Nice! We need to check if it's working on iOS. @StuffJoy could you try on your mac emulator?

epou:
We met with @StuffJoy yesterday, and iOS needs further investigation. I would proceed with this pull request only for android and continue with issue #59 opened, pending to double check iOS.

PR #124, opened 2024-04-30
Remove My Reports -> Map

DigitalSeneca:
fixes #99 

| My reports | Opened report |
| ------------- | ------------- |
| image | image |

Issue #99, opened 2024-03-25
Remove My Reports -> Map


epou:
Keep only My Reports -> List

PR #128, opened 2024-05-10
MyReportsPage with 3 tabs

### LINKED ISSUE #109: Add 3 tab (adult,bite,sites) in My Report list
Example:

https://miro.medium.com/v2/resize:fit:441/1*XzhY38-Dv4SpcF1AglloeQ.png

DigitalSeneca:
fixes #109 and #99

replaces PR #124 

3_tabs


lib/models/report.dart
```diff
@@ -108,8 +108,14 @@ class Report {
     country = json['country'];
     nuts3 = json['nuts_3'];
     nuts2 = json['nuts_2'];
+  }
+
+  static Future<Report> fromJsonAsync(Map<dynamic, dynamic> json) async{
+    var report = Report._internal(json);
+
+    await report._initializeDisplayCity();
```
DigitalSeneca, line 116: By having it async it has time to load the displayCity. Before this, sometimes it was null randomly

lib/models/report.dart
```diff
@@ -161,6 +168,14 @@ class Report {
     data['app_language'] = app_language;
     return data;
   }
+
+  LatLng getLocation() {
```
epou, line 172: There's actually a third scenario we need to consider with location_choice: sometimes it's marked as 'missing' for old reports that didn't include location data. In those cases, it might be cool to just return null.
It should be interesting to force location_choice == current , location_choice == selected and else null.

lib/pages/my_reports_pages/my_reports_page.dart
```diff
@@ -1,925 +1,84 @@
...
+          backgroundColor: Colors.white,
+          bottom: TabBar(
+            tabs: [
+              Tab(
+                icon: Image.asset(
+                  'assets/img/ic_my_reports_mosquito.png',
+                  width: 24,
+                  height: 24,
+                  filterQuality: FilterQuality.high),
+                text: MyLocalizations.of(context, 'single_mosquito') ?? 'Mosquito'),
```
epou, line 52: Isn't there already a fallback to english for translations?

lib/models/report.dart
DigitalSeneca: Okey! I've added the third scenario, but should we return null or (0,0) like when gps is disabled? If we return null, I'd have to disable the minimap and location coordinates inside of "My Reports"

Not related with your comment, but related with this method: I also considered using Position from geolocator instead of LatLng from GoogleMaps as we talked, but there were 2 drawbacks:
- Position requires the method to be async instead of sync
- Position required more parameters to be set (altitude, timestamp...), while LatLng is just latitude and longitude, which is what we need

epou: Conceptually speaking, in my opinion, it makes sense to return null instead of latlon 0,0.

lib/pages/my_reports_pages/my_reports_page.dart
DigitalSeneca: yes, but https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/f71905e91e2fb0084f83387bb5328930baa0e4a6/lib/utils/MyLocalizations.dart#L46 has a datatype of String? instead of String so it confused me, I've modified it to not be nullable, as it shouldn't be

lib/models/report.dart
DigitalSeneca: I've made it nullable and I think it looks good. If location is null, I just don't show the minimap, coordinates and displayCity like in this screenshot:

image

epou:
I've seen you've request my review... but there's still pending comments from my previous one

DigitalSeneca:
> I've seen you've request my review... but there's still pending comments from my previous one

I think I already answered all the previous comments about the third scenario with location_choice and the fallback to English. At least I am not seeing other unanswered comments

epou (changes requested):

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -1,98 +1,549 @@
...
+                  ),
+                title: Text(getTitle(context, reports[index])),
+                subtitle: RichText(
+                  text: TextSpan(
+                    children: [
+                      WidgetSpan(
+                        child: Icon(Icons.location_on, size: 16),
+                      ),
+                      TextSpan(
+                        text: ' ${reports[index].displayCity ?? ''}\n',
```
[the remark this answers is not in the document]
DigitalSeneca: By default subtitle is a one line. By adding \n, displayCity goes in the first line and creation_time goes in the second line, while both still belong to the subtitle

lib/utils/MyLocalizations.dart
```diff
@@ -43,7 +43,7 @@ class MyLocalizations {
     return _englishValues[key] ?? '';
   }
 
-  static String? of(BuildContext context, String? key) {
+  static String of(BuildContext context, String? key) {
```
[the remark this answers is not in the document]
DigitalSeneca: yeah, this comes from the upgrade to flutter 3.0.0 with null safety. Before making it non nullable here we have to modify other variables to be non nullable

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -1,98 +1,528 @@
...
+          itemCount: reports.length,
+          itemBuilder: (context, index) {
+            return Card(
+              shape: RoundedRectangleBorder(
+                borderRadius: BorderRadius.circular(10),
+              ),
+              elevation: 4.0,
+              child: ListTile(    
+                leading: reports[index].type == 'bite' ? null :
+                  CircleAvatar(
```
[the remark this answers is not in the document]
DigitalSeneca: Sure, like this?
image

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -1,98 +1,557 @@
...
+        return '${report.responses![0]?.answer_value ?? ''} bites';
+      case 'site':
+        return MyLocalizations.of(context, report.responses![0]?.answer ?? '');
+      default:
+        return '';
+    }
+  }
 
   @override
   Widget build(BuildContext context) {
```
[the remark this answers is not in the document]
DigitalSeneca: I've added again the code that I removed accidentally after the refactor, to sort it in descending order

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -1,98 +1,545 @@
...
+  }
+
+  String getTitle(BuildContext context, Report report){
+    switch(report.type) {
+      case 'adult':
+        var response_question_6 = report.responses?.firstWhere(
+          (element) => element!.question == 'question_6');
+
+        if (response_question_6 == null){
+          return '';
```
epou, line 43: return 'Non identified'. Once approved the code of the new 'adult report report flow', this question will stop being answered, and all reports will rely on this 'if' statement.

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -1,98 +1,545 @@
...
+        }
+
+        if (response_question_6.answer == 'question_6_answer_64'){
+          return 'Non identified';
+        }
+
+        return MyLocalizations.of(context, response_question_6.answer);
```
epou, line 50: In case 'MyLocalizations.of(context, response_question_6.answer)' does return empty or null, could you return the 'response_question_6.answer' without being translated?

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -1,98 +1,545 @@
...
+      case 'bite':
+        var response_question_1 = report.responses?.firstWhere(
+          (element) => element!.question == 'question_1');
+
+        var numBites = response_question_1?.answer_value;
+        if (numBites == null){
+          return '';
```
epou, line 57: return 0 ?

lib/pages/my_reports_pages/components/reports_list_widget.dart
```diff
@@ -1,98 +1,545 @@
...
+        }
+        if (numBites == '1'){
+          return '$numBites bite';
```
epou, line 60: You can use the translation of the string 'single_bite' for these, will be wrong in determining singular/plural forms thought.

`return "$numBites MyLocalizations.of(context, 'single_bite')"`

lib/pages/my_reports_pages/components/reports_list_widget.dart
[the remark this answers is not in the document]
epou: Awesome! Thanks :)

lib/pages/my_reports_pages/components/reports_list_widget.dart
DigitalSeneca: will we have a translation string also for the plural? so that is not hardcoded for plural and for 0

lib/pages/my_reports_pages/components/reports_list_widget.dart
DigitalSeneca: I've added a check in case it's an empty string (If the answer is null, the translation will come as empty string)

lib/pages/my_reports_pages/components/reports_list_widget.dart
epou: I'm creating a new string in localize for it: 'plural_bite'

epou: Just make sure to lowercase the result in both cases:
OK: 3 bites
OK: 1 bite
KO: 3 Bites

epou: Already created in localise.biz

epou (commented):
Ok from my end. Just missing to add the new translations for 'plural_bite'

lib/pages/my_reports_pages/components/reports_list_widget.dart
DigitalSeneca: I've added it now, in lowercase and also downloaded the strings from localise.biz

assets/language/mk_MK.json
epou, line 1: Remove this one from the pull-request please

DigitalSeneca: okey, I've removed the changes of mk_MK.json

epou:
> What do you mean? I've added `plural_bite` to MyReportsPage and also downloaded from localise.biz

Resolved in the commit https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/128/commits/c027bb5ed31952c9d50be5cb02cb51a8112afc9f

epou (approved):

Issue #105, opened 2024-03-25
Make adult reporting faster


epou:
We have now more than 5 click in until the user can submit the final report.

Flow:
1. Click on Report Adult
2. Open camera with gallery options (preview last images).
3. User shots image/select images from gallery. (maximum 3)
4. Form:
    - Moment of the interaction (now, last night, last midday, last morning): use icons only
    - Where you bitten? (toggle switch)
    - Map (if GPS enable set, if user click on the map -> 'selected')
    - Environment: indoors, vehicle, outdoors: use icons only
    - Notes:

Something similar to:
![Screenshot from 2024-03-25 09-59-27](https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/assets/26668968/3aa1fcee-25d5-4b4d-987c-601f33a7d031)

Issue #103, opened 2024-03-25
Adult report must contain photos


epou:
Force adult reports to have images.

On clicking "Report Adult" go directly to Camera with Gallery options (see WhatsApp style, where last images are previewed in top of the shutter button)

Add Badge text on top: "Only one individual"

PR #129, opened 2024-05-15
New adult report workflow

### LINKED ISSUE #17: Make app open to any adult mosquito species
Simplify the adult report flow such that the app is no longer focused only on the current 5 targets. This is the first stage in a broader set of changes that we will want to implement for the 2024 season. In this initial stage the idea is to make a very simple change that does not change the data structure. Simply changing the flow of the report screens such that the participant is not pushed away from reporting non-target species.

DigitalSeneca:
fixes #17 #103 #105


epou:
By the way, it's important only to allow posting images from the gallery or camera and ensure video is not allowed.

DigitalSeneca:
> By the way, it's important only to allow posting images from the gallery or camera and ensure video is not allowed.

This is still work in progress, but for now I have this (screenshot) that is opened when you click "Report mosquito" and you can both take a picture or select from gallery. From the emulator I am not able to take a video, I'll also test it from a phone before marking it as ready

image

I am using the [whatsapp_camera package](https://pub.dev/packages/whatsapp_camera) which shows a GIF in whatsapp style, but I did not find the way to show "recent photos" in the bottom, only to select from gallery with the buttom in top right corner

DigitalSeneca:
closed in favor of #131

PR #131, opened 2024-05-23
Directly open camera + Open app to any adult mosquito species

DigitalSeneca:
### Description of changes
- #103
  - Directly open camera
  - Force adult reports to have images
- #17
  - Remove "What type of mosquito it was" and thorax,abdomen,leg...?

| GIF 1 | GIF 2 |
| directly_open_camera | directly_open_camera_2 |


lib/pages/forms_pages/camera_page.dart
```diff
@@ -0,0 +1,45 @@
...
+    if (photos != null && photos.length >= 1 && photos.length <= 3) {
+      await Navigator.push(
+        context,
+        MaterialPageRoute(
+          builder: (context) => AdultReportPage(photos: photos),
+        ),
+      );
+    }
+    Navigator.pop(context);
+  }
```
epou, line 33: Going to the previous page in case of selecting more than three images without any explanation can be confusing for users. To address this, consider the following solutions:

a) Limit the number of selectable images in the gallery FilePicker to a maximum of three.
b) Configure WhatsAppCamera with `multiple: false`, so users can only add images one by one.
c) Display a dialog when more than three images are selected, prompting the user to unselect the excess images.

lib/pages/forms_pages/components/add_photo_button_widget.dart
```diff
@@ -4,181 +4,128 @@ import 'dart:math';
...
-    var response = await http.get(Uri.parse(imageUrl));
-    // write bodyBytes received in response to file.
-    await file.writeAsBytes(response.bodyBytes);
-    // now return the file which is created with random name in
-    // temporary directory and image bytes from response is written to // that file.
-    return file;
+  Widget squareToAddPhoto() {
+    return GestureDetector(
+      onTap: () {
+        _chooseTypeImage();
```
epou, line 71: Could you use the whatsapp_camera again here?

lib/pages/main/main_vc.dart
```diff
@@ -11,6 +11,7 @@ import 'package:mosquito_alert_app/models/notification.dart';
 import 'package:mosquito_alert_app/pages/forms_pages/adult_report_page.dart';
```
epou, line 11: So then, I guess this import is no longer needed

lib/pages/forms_pages/adult_report_page.dart
```diff
@@ -299,6 +189,8 @@ class _AdultReportPageState extends State<AdultReportPage> {
     if (widget.editReport != null) {
       widget.loadData!();
     }
+    Navigator.pop(context);
+    Navigator.pop(context);
```
epou, line 193: Duplicated?

lib/pages/forms_pages/breeding_report_page.dart
epou, line 1: Could you replace the use of `_chooseTypeImage()` for the new whatsapp camera?

epou:
(Pending: ensure no video can be selected from the gallery)

lib/pages/forms_pages/adult_report_page.dart
DigitalSeneca: In this case we need 2 Navigator.pop() to go back to the homepage. Since the camera is open, the first pop() goes to the CameraPage but with the camera closed, and the second one goes to homepage. Since CameraPage without the camera open is not useful, I just skip it

epou: understood, thank's for the clarification :smile:

lib/pages/forms_pages/camera_page.dart
DigitalSeneca: Let's go with "multiple: false", seems the most straightforward solution. If they need to add more photos they can do it in the next screen

epou:
I was wondering if the 'camera flip function' (button) could be disabled. This change aims to reduce the likelihood of users taking selfies.

DigitalSeneca:
> I was wondering if the 'camera flip function' (button) could be disabled. This change aims to reduce the likelihood of users taking selfies.

I don't see this option unfortunately

epou:
@aescobarr , we can confirm that videos are discarded from the gallery and that the camera only allows taking pictures.

epou:
(let me know once the pull request is ready for review)

DigitalSeneca:
closed in favor of #165

PR #132, opened 2024-05-28
Fix app not opening when region is different

### LINKED ISSUE #112: ANR: "There is only the white screen with the logo"
**Describe the bug**
"There is only the white screen with the logo. I tried to delete and re-install the app (several times), but the problem remains. And it is not just my phone, a colleague of mine had the same problem."

**Expected behavior**
It should open the app normally

**Smartphone (please complete the following information):**
Happened to multiple users
- Galaxy S10e, model number SM-G970F/DS. Android 12
- Galaxy S10, model number SM-G973F/DS. Android 12

Many users could not reproduce this bug on Android 9, 12, 13 and 14 -> They reported it's working normally for them

**Additional context**
It's probably the same error or related to [Google Developer Console: dalvik.system.BaseDexClassLoader.findClass](https://play.google.com/console/u/0/developers/6847921171114158955/app/4973233962343162851/vitals/crashes/db9355fa2f87e85199eb3ab58da965e3/details?days=28&versionCode=1717%2C1708&isUserPerceived=true)

DigitalSeneca:
fixes #112


lib/utils/MyLocalizations.dart
```diff
@@ -14,15 +14,23 @@ class MyLocalizations {
...
+    var jsonContent = await rootBundle.loadString('assets/language/en.json');
     _englishValues = json.decode(jsonContent);
   }
 
   static Future<MyLocalizations> loadTranslations(Locale locale) async {
     var appTranslations = MyLocalizations(locale);
-    var jsonContent = await rootBundle.loadString(
-      'assets/language/${locale.languageCode}_${locale.countryCode}.json'
-    );
+    late var jsonContent;
```
epou, line 23: I would use String? so it init as null
`String? jsonContent;`

lib/utils/MyLocalizations.dart
```diff
@@ -14,15 +14,23 @@ class MyLocalizations {
...
+      jsonContent = await rootBundle.loadString(
+        'assets/language/${locale.languageCode}_${locale.countryCode}.json'
+      );
+    } catch (_) {
+      jsonContent = await rootBundle.loadString(
+        'assets/language/${locale.languageCode}.json'
+      );
+    }
+
     _localisedValues = json.decode(jsonContent);
```
epou, line 34: In case jsonContent has not been loaded. pass {} to `_localisedValues`

```dart
_localisedValues = json.decode(jsonContent ?? '{}')
```

epou (commented):
Added some comments. If no breaking changes, it'd be ideal to merge against 'master', and do a fast release to solve this bug.

lib/utils/MyLocalizations.dart
DigitalSeneca: I changed the variable declaration "String? jsonContent;" to be nullable, but when we call "await rootBundle.loadString()", it returns a non-nullable string so then your code proposal would return this warning

"The operand can't be null, so the condition is always 'true'.
Remove the condition.dartunnecessary_null_comparison"

lib/utils/MyLocalizations.dart
```diff
@@ -20,9 +20,29 @@ class MyLocalizations {
...
+    final regionFilename = 'assets/language/${locale.languageCode}_${locale.countryCode}.json';
+
+    try {
+      // Try loading the region-specific file
+      jsonContent = await rootBundle.loadString(regionFilename);
+    } catch (_) {
+      // If not found, load the language from any region
+      final supportedLanguages = ['bg_BG', 'bn_BD', 'ca_ES', 'de_DE', 'el_GR', 'en_US', 'es_ES', 'es_UY', 'eu_ES',
+        'fr_FR', 'gl_ES', 'hr_HR', 'hu_HU', 'it_IT', 'lb_LU', 'mk_MK', 'nl_NL', 'pt_PT', 'ro_RO', 'sl_SI',
+        'sq_AL', 'sr_RS', 'sv_SE', 'tr_TR',
```
DigitalSeneca, line 33: I think it's not ideal or good to have the codes here because they are also hardcoded in settings_page.dart. However, loading the files using the directory was not working as `assets/language` was not being detected (I don't know why, since it's added to pubspec.yaml and the directory should be correct)

DigitalSeneca:
> Added some comments. If no breaking changes, it'd be ideal to merge against 'master', and do a fast release to solve this bug.

I can't directly change the target of this PR to master because it has the changes from v34, but once it's approved I can create a new PR targeted to master with just those changes

PR #137, opened 2024-06-10
Main view has only 3 buttons for reporting

### LINKED ISSUE #100: Main view should only be 3 buttons for reporting
Change Heading to a single title (translate): `What would you like to report?`

3 rounded colored buttons:
- Adult: color `DFD458` 40%
- Bite: color `D28A73` 40%
- Breeding site: color `7D9393` 40%

Move "My data" to a the navigationdrawer

Screenshot from 2024-03-25 09-43-43

DigitalSeneca:
fixes #100

- I would not merge it until the PR with the drawer is also ready, because the "My Reports" button would be gone otherwise
- I would need to receive the translation string for "What's in your surroundings?" or "What would you like to report?"

image

### DIFF
--- a/lib/pages/main/components/custom_card_widget.dart
+++ b/lib/pages/main/components/custom_card_widget.dart
@@ -1,51 +1,40 @@
 import 'package:flutter/material.dart';
-import 'package:mosquito_alert_app/utils/style.dart';
+import 'package:mosquito_alert_app/utils/MyLocalizations.dart';
 
 class CustomCard extends StatelessWidget {
-  final String? img, title;
-  final bool? selected, disabled;
-  CustomCard({this.img, this.title, this.selected, this.disabled});
+  final String text, image_path, color;
+  final dynamic reportFunction;
+  CustomCard({required this.text, required this.image_path, required this.color, required this.reportFunction});
 
   @override
   Widget build(BuildContext context) {
-    return Opacity(
-      opacity: disabled == true ? 0.5 : 1,
+    return Material(
+      elevation: 5.0,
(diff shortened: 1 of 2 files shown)


lib/pages/main/main_vc.dart
```diff
@@ -180,202 +177,39 @@ class _MainVCState extends State<MainVC> {
...
+                                                MyLocalizations.of(context, 'what_to_do_txt'),
+                                                fontSize: 16),
+                                          ]),
+                                      )                                      
+                                    ]),
+                                  SizedBox(height: 25),
+                                  CustomCard(
+                                    text: 'single_mosquito',
+                                    image_path: 'assets/img/ic_mosquito_report.png',
+                                    color: '40DFD458',
```
DigitalSeneca, line 197: First two characters indicate the opacity (in this case, 40) and the rest of characters indicate the color

lib/pages/main/components/custom_card_widget.dart
DigitalSeneca, line 1: I have removed the code in CustomCard entirely and added new code, since this was only used in HomePage

epou:
Great! Could you add some left margin to the images so that it matches the thickness of the top and bottom margins?

DigitalSeneca:
> Great! Could you add some left margin to the images so that it matches the thickness of the top and bottom margins?

I've added a left padding of 5.0 and updated the screenshot in the screenshot. Now the image matched the thickness better, but I can still update it as needed

DigitalSeneca:
closed in favor of #143 (stacked pull request with target to drawer instead of v34)

Issue #52, opened 2023-12-22
Show public map in the app


epou:
As an option inside the NavigationDrawer.

Webview: map.mosquitoalert.com/<lang> {en,ca,es}

Centered in the current user position (will need to adapt the frontend code)

PR #142, opened 2024-06-12
Display public map in a webview

DigitalSeneca:
fixes https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/issues/52

> Centered in the current user position (will need to adapt the frontend code)

We need to discuss how to implement this, maybe it makes sense to do it via API

image


epou:
> We need to discuss how to implement this, maybe it makes sense to do it via API

The new version of the map (to be released in the next weeks), will accept lat/lon/zoom parameters + have geolocation button. Nothing to worry about for this pull request, we ca change it the URL params once the new map is in prod.

epou (approved):

PR #145, opened 2024-06-13
new info page

DigitalSeneca:


lib/pages/settings_pages/info_page.dart
DigitalSeneca, line 1: That's the same as settings page but only with info options

PR #152, opened 2024-06-19
Changed 'My reports' bite tab icon

epou:


DigitalSeneca (commented):
The icon itself looks good :)
but it seems smaller compared to the mosquito icon on its left
image

epou:
Could you try again with this? (removed transparent margin)

bite icon

DigitalSeneca:
> Could you try again with this? (removed transparent margin)
> 
> bite icon

still looks small, can you try without the transparent margin but still keeping it a square? I'm thinking the problem could be the new image being a rectangle

epou:
Well.... in that case maybe it's worth trying with a different image. How about this?

bite2

DigitalSeneca:
This one looks much better! Do you want me to push a commit with it or are you going to do it?

image

epou:
Could you commit it for me? I'm not at the office right now

DigitalSeneca:
done!

PR #154, opened 2024-06-19
Header - Reduce size of the notification bell

DigitalSeneca:
image


epou:
Could you try removing `position: badges.BadgePosition.topEnd(top: 2, end: 2),` to see if the red circle is nearer to the icon?

DigitalSeneca:
> Could you try removing `position: badges.BadgePosition.topEnd(top: 2, end: 2),` to see if the red circle is nearer to the icon?

Actually, instead of removing it we have to increase the numbers. That's because it acts kind of as a margin, starting from the top right, like in the screenshot
image

epou (approved):
Much better! Thanks @DigitalSeneca :)

PR #157, opened 2024-06-20
Remove version_control.dart dribba

### LINKED ISSUE #146: Remove version_control.dart
We are no longer working with the company that provide the API endpoint used to check the version + this is prone to errors.

DigitalSeneca:
fixes #146 

To me it seems that there are no side affects. I'm not sure if I should test something specific


epou (commented):
Looks good to me, shouldn't be any side effect... Maybe it would be a good idea to rename main_vc.dart to main.dart?

DigitalSeneca:
> Looks good to me, shouldn't be any side effect... Maybe it would be a good idea to rename main_vc.dart to main.dart?

Then we have 2 main.dart files. Maybe it's more appropriate to name it something related to the drawer/header, since that's the most contents of this file

epou:
Yes, just choose a name that makes sense and makes you feel comfortable. Feel free

DigitalSeneca:
How does `drawer_and_header.dart` sound?

epou:
Go for it! Sounds good to me :+1:

PR #168, opened 2024-06-25
Add text to provide more guidance on attaching photo

DigitalSeneca:
I'd need some translation strings to follow feedback given by users

image


DigitalSeneca:
Should we close this in favor of #181 ?

epou:
I like how #181 looks, but I think the text there should be shorter. I would merge this PR and continue with the other text in #181, so at the end we will end having both

DigitalSeneca:
Ah okay, understood. I've updated the 2 screenshots in the PR description. For the breeding report, should I use a different text, leave it empty...?

epou:
Only for adults, nothing to display for breeding sites.

DigitalSeneca:
perfect, it's showing only for adults now

lib/pages/forms_pages/components/add_photo_button_widget.dart
```diff
@@ -14,8 +14,9 @@ class AddPhotoButton extends StatefulWidget {
   final bool isEditing;
   final bool photoRequired;
   final Function _atLeastOnePhotoAttached;
+  final bool isAdultReport;
 
-  AddPhotoButton(this.isEditing, this.photoRequired, this._atLeastOnePhotoAttached);
+  AddPhotoButton(this.isEditing, this.photoRequired, this._atLeastOnePhotoAttached, this.isAdultReport);
```
epou, line 19: I think it's better to make the AddPhotoButton class agnostic of the type of report. I propose using a string variable like `this.subtitle` or similar, and only show the text if that variable is not null or empty.

DigitalSeneca: you're right, if modified it. I'll do the same change for the other PR too

lib/pages/forms_pages/adult_report_page.dart
```diff
@@ -75,7 +75,7 @@ class _AdultReportPageState extends State<AdultReportPage> {
     _pagesController = PageController();
     index = 0.0;
     _initialformsRepot = [
-      AddPhotoButton(true, true, _checkAtLeastOnePhotoAttached),
+      AddPhotoButton(true, true, _checkAtLeastOnePhotoAttached, 'ensure_single_mosquito_photos'),
```
epou, line 78: Should here be using the already translated string instead of the string id? I mean directly the `MyLocalizations.of(context, 'ensure_single_mosquito_photos')`

DigitalSeneca: I tried it like that at first, but then get an exception
image

epou: so then, maybe it make more sense to just simply rename subtitle to subtitled or similar

DigitalSeneca: maybe `subtitleKey` makes sense since it's the key to be translated inside the `AddPhotoButton` widget

epou (approved):

PR #178, opened 2024-07-01
Fix null error when exiting camera without selecting a photo

DigitalSeneca:
| image | image |


epou (approved):
Good catch!

PR #179, opened 2024-07-01
Move gallery button to bottom right

### LINKED ISSUE #17: Make app open to any adult mosquito species

DigitalSeneca:
fixes #17 
> could the camera layout be adjusted to place the gallery button next to the shutter button (on its right) instead of in the top-right corner?

image


epou (approved):
Bravo!

PR #181, opened 2024-07-01
Add informational text on top of the camera view

DigitalSeneca:
image


epou:
For adult reports only, use localise text id: `one_mosquito_reminder_badge`

Change text color to white
Can you reduce the top margin to look like it is placed just after the header? (considering the header to be the invisible full-width box where the X symbol is placed)

DigitalSeneca:
What text should I show in the camera when reporting a breeding site? Or should the text only appear for reporting adults?

I have applied some changes:
image

epou:
Amazing! Nothing to show (at the moment) when reporting breeding sites, only show the text when reporting adults.

DigitalSeneca:
Now it's only displayed for adult reports

lib/pages/forms_pages/components/whatsapp_camera.dart/camera_whatsapp.dart
```diff
@@ -139,7 +141,7 @@ class WhatsappCamera extends StatefulWidget {
   ///
   ///```
   ///
-  const WhatsappCamera({key, this.multiple = true});
+  const WhatsappCamera({key, this.multiple = true, required this.isAdultReport});
```
epou, line 144: Wouldn't it be simpler if instead of adding a boolean using a string instead? 

something like `this.InfobadgeText` and only show is that variable is not null or empty? i think it's more scalable

lib/pages/forms_pages/components/whatsapp_camera.dart/camera_whatsapp.dart
```diff
@@ -196,6 +198,28 @@ class _WhatsappCameraState extends State<WhatsappCamera>
...
+          Visibility(
+            visible: widget.isAdultReport,
+            child:Align(
+              alignment: Alignment.topCenter,
+              child: Padding(
+                padding: const EdgeInsets.only(top: 70),
+                child: Container(
+                  width: 0.5 * MediaQuery.of(context).size.width,
+                  decoration: BoxDecoration(
+                    color: Colors.orange,
```
epou, line 210: Could you try with `Colors.orange` with a degree of shade (`Color.orange.shade300` or `Color.orange.shade200` maybe??)? If you think the UI is improved go ahead

lib/pages/forms_pages/components/whatsapp_camera.dart/camera_whatsapp.dart
epou: So then, the text `MyLocalizations.of(context, 'one_mosquito_reminder_badge')` will be placed in adult_report_page.dart, which makes more sense

lib/pages/forms_pages/components/whatsapp_camera.dart/camera_whatsapp.dart
DigitalSeneca: it looks lighter, let's go ahead

lib/pages/forms_pages/breeding_report_page.dart
```diff
@@ -94,7 +94,7 @@ class _BreedingReportPageState extends State<BreedingReportPage> {
         true,
         goNextPage,
         'assets/img/bottoms/breeding_1.png'),
-      AddPhotoButton(true, true, _checkAtLeastOnePhotoAttached),
+      AddPhotoButton(true, true, _checkAtLeastOnePhotoAttached, 'site'),
```
epou, line 97: Shouldn't be 'site' replaced for null here?

lib/pages/forms_pages/adult_report_page.dart
```diff
@@ -75,7 +75,7 @@ class _AdultReportPageState extends State<AdultReportPage> {
     _pagesController = PageController();
     index = 0.0;
     _initialformsRepot = [
-      AddPhotoButton(true, true, _checkAtLeastOnePhotoAttached),
+      AddPhotoButton(true, true, _checkAtLeastOnePhotoAttached, 'adult'),
```
epou, line 78: Shouldn't be 'adult' replaced with `MyLocalizations.of(context, 'one_mosquito_reminder_badge')` here so AddPhotoButton class is agnostic of the report type?

DigitalSeneca: If modified it to be type agnostic. Same as the other PR, I have to pass the translation key instead of the value directly, and then from `AppPhotoButton` again pass it to `WhatsappCamera` to build the widget

epou (approved):

PR #183, opened 2024-07-05
Improve Settings -> Autotag

### LINKED ISSUE #134: Improve Settings -> Autotag
Complaints that the current way of adding and removing tags is not straightforward.

Here's my proposal for change:
Folded:
autotag_folded
Unfolded:
autotag_unfolded

To take into consideration:
- badge counter should disappear when 0 (no tags added)

Interesting libraries: 
- https://pub.dev/packages/textfield_tags

DigitalSeneca:
fixes #134 

autotag_orange


DigitalSeneca:
@epou 
1. Do we have translation strings for 'Auto-tag reports' and ''Enter tag...'? (See GIF in the description)
2. The API endpoint accepts the hashtag as a string, so I'm concatenating it into a string, ie ['tag1', 'tag2', 'tag3] -> 'tag1, tag2, tag3'. If you're changing the API endpoint let me know so that I can directly send the list instead

epou:
> 1. Do we have translation strings for 'Auto-tag reports' and ''Enter tag...'? (See GIF in the description)
You can use the following string_ids:

- Title: `auto_tagging_settings_title`
- Subtitle (only show when unfolded): `enable_auto_hashtag_text`
- Placeholder: `auto_tagging_settings_placeholder`

> 2. The API endpoint accepts the hashtag as a string, so I'm concatenating it into a string, ie ['tag1', 'tag2', 'tag3] -> 'tag1, tag2, tag3'. If you're changing the API endpoint let me know so that I can directly send the list instead

As far as I know, the current API should get the tags from the 'notes' field, and should be a text-like field. So then, the tags should be appended to the notes field automatically in the form of a text list where each element begins with a `#`. like: `#tag1, #tag2, #tag3`. In the case the user has added a note, it should be appended at the end, example: `Here the user notes talking about whatever. #tag1, #tag2, #tag3` @aescobarr could you confirm that?

aescobarr:
> > 2. The API endpoint accepts the hashtag as a string, so I'm concatenating it into a string, ie ['tag1', 'tag2', 'tag3] -> 'tag1, tag2, tag3'. If you're changing the API endpoint let me know so that I can directly send the list instead
> 
> As far as I know, the current API should get the tags from the 'notes' field, and should be a text-like field. So then, the tags should be appended to the notes field automatically in the form of a text list where each element begins with a `#`. like: `#tag1, #tag2, #tag3`. In the case the user has added a note, it should be appended at the end, example: `Here the user notes talking about whatever. #tag1, #tag2, #tag3` @aescobarr could you confirm that?

Yes, that should be the case. The hashtags have to be somewhere inside the fields note, it does not matter where but at the end is nice and easily readable when looking at the database content. About separating multiple hashes with commas, it's also ok and looks nice, but it's not necessary either, spaces should be enough as long as the different hashes are different substrings inside the notes field.

The thing that's really important is that each hashtag starts with the '#' character. Never remove this when adding the tags at the end of the string, since if there is no '#' we can't distinguish between somebody using the word 'tag' in the notes and the hashtag '#tag'.

DigitalSeneca:
Thank you for the explanation, I've implemented it:
image

lib/api/api.dart
DigitalSeneca, line 1: The change in this file is to go from appending a single hashtag to the notes, to allowing multiple hashtags, all of them prepended by the hashtag symbol -> #

lib/pages/settings_pages/components/hashtag.dart
DigitalSeneca, line 1: Most of the code of this file is an official example from https://pub.dev/packages/textfield_tags
I made modifications to the implementation to adapt the style and to modify the `hashtags` values on `UserManager` every time the user adds or removes a tag
I also added the `CircularProgressIndicator` to show while I load the values of `initialTags`

lib/pages/settings_pages/settings_page.dart
DigitalSeneca, line 1: I removed all the code related to the tags, and wrote the new logic in hashatg.dart

lib/pages/settings_pages/components/hashtag.dart
```diff
@@ -0,0 +1,156 @@
...
+                          width: 3.0,
+                        ),
+                      ),
+                      focusedBorder: const OutlineInputBorder(
+                        borderSide: BorderSide(
+                          color: Colors.grey,
+                          width: 3.0,
+                        ),
+                      ),
+                      hintText: inputFieldValues.tags.isNotEmpty ? '' : 'Enter tag...',
```
epou, line 82: localise: Use string_id: `auto_tagging_settings_placeholder`

lib/pages/settings_pages/settings_page.dart
```diff
@@ -134,17 +125,17 @@ class _SettingsPageState extends State<SettingsPage> {
...
-              height: 60,
+            Container(
+              padding: EdgeInsets.only(bottom: 12.0, top: 12.0),
+              decoration: BoxDecoration(
+                borderRadius: BorderRadius.circular(5),
+                color: Colors.white,
+                border: Border.all(color: Colors.black.withOpacity(0.1))
+              ),
+              child: ListTile(
+                title: Style.body('Auto-tag reports'),
```
epou, line 136: localise: Use string_id: `auto_tagging_settings_title`

epou:
Is there a way of making it foldable + adding counter? (see image in issue #134 )

lib/pages/settings_pages/settings_page.dart
DigitalSeneca: I've added the 2 localise string_id

DigitalSeneca:
> Is there a way of making it foldable + adding counter? (see image in issue #134 )

Yes, in the first version I made it work. Now I've just pushed this one where it's foldable. Only the counter is missing, which should need a similar implementation as the notification

new_autotag

epou:
Nice! So then, only the counter is missing (only show if not 0). I was wondering if you could add the text `enable_auto_hashtag_text` when the section is unfolded. It's the description of what auto-tagging means.

DigitalSeneca:
I've added the counter now too
new_autotag_counter

> I was wondering if you could add the text enable_auto_hashtag_text when the section is unfolded. It's the description of what auto-tagging means.

I just added this description too. It makes the tile bigger but it's very descriptive:
image

epou:
@DigitalSeneca you've used the wrong string_id for the description, it should be `enable_auto_hashtag_text`

DigitalSeneca:
sorry I took your string and then messed it up when getting the same font color from the background tracking

lib/pages/settings_pages/settings_page.dart
```diff
@@ -134,18 +140,57 @@ class _SettingsPageState extends State<SettingsPage> {
...
+              ),
+              child: Row(
+                children: [
+                  Expanded(
+                    child: ExpansionTile(
+                      title: Row(
+                        children: [
+                            Text(
+                              MyLocalizations.of(context, 'auto_tagging_settings_title'),
+                            ),
```
epou, line 157: Add `Spacer()` to push the counter to the right

lib/utils/UserManager.dart
```diff
@@ -204,9 +204,9 @@ class UserManager {
     return prefs.getStringList('imagesList');
   }
 
-  static Future<String?> getHashtag() async {
+  static Future<List<String>?> getHashtags() async {
     var prefs = await SharedPreferences.getInstance();
-    return prefs.getString('hashtag');
+    return prefs.getStringList('hashtags');
```
epou, line 209: It occurred to me that users who have already enabled the auto-tagging function will lose their tags due to the renaming of 'hashtag' to 'hashtags'.

To prevent this, we should consider applying a migration during app initialization. I was thinking on something like:
```dart
Future<void> migrateHashtagToHashtags() async {
  final prefs = await SharedPreferences.getInstance();

  // Check if the old variable exists
  if (prefs.containsKey('hashtag')) {
    // Read the old variable's value
    String? oldHashtag = prefs.getString('hashtag');

    // Save the value to the new variable as a list
    if (oldHashtag != null) {
      await prefs.setStringList('hashtags', [oldHashtag]);
    }

    // Remove the old variable
    await prefs.remove('hashtag');
  }
}
```

epou: Ei, Santi just told me that the users are currently setting the tags by specifying '#' as prefix. We need to take that into account for the migration... May be applying a extraction function in the function in the previous comment (using regex):
```dart
// Function to extract hashtags
  final RegExp hashtagRegExp = RegExp(r'#(\w+)');
  return hashtagRegExp
      .allMatches(input)
      .map((match) => match.group(1)!)
      .toList();
}
```

or using split(' ').startswith('#') ....

DigitalSeneca: I've added this and wrote a test with the 2 cases to check it's migrated as expected
image

lib/pages/settings_pages/settings_page.dart
DigitalSeneca: done
image

epou (approved):

PR #189, opened 2024-07-13
Request users to submit a first review after they submit a 3rd, 4th, (6, 9, 12, 15...) report

### LINKED ISSUE #148: Ask users for in-app-reviews
https://pub.dev/packages/in_app_review

Display when:
- The user has not reviewed the app yet (any of the following):
    - Immediately after the 3RD uploaded report. (IQR -> Q3)
    - Immediately after the 4TH uploaded report.
    - The last 'review request' was = 3 reports ago || 2 week ago

links
- https://medium.com/@erickogaro1998/boost-your-apps-ratings-and-user-engagement-with-in-app-reviews-a-complete-flutter-guide-b3882e515ab5
- https://30dayscoding.com/blog/adding-in-app-review-prompts-to-flutter-apps
- inspiration https://github.com/Skyost/RateMyApp
- Alternative: https://pub.dev/packages/app_review

DigitalSeneca:
fixes #148 

- According to https://pub.dev/packages/in_app_review#testing-read-carefully , the only way to test it is by building an appbundle and uploading it to Play Store as internal test
- I tested that once a review has been submitted, the app does not request again for a review after 6, 9, 12... reviews

| First Header  | Second Header | Result in PlayStore |
| ------------- | ------------- | ------------- |
| image | image | image |

### DIFF
--- a/lib/utils/Utils.dart
+++ b/lib/utils/Utils.dart
@@ -1203,8 +1203,6 @@ class Utils {
     var myReports = await ApiSingleton().getReportsList();
     var numReports = myReports.length;
 
-    if (numReports < 3) { return; }
-
     final now = DateTime.now().millisecondsSinceEpoch;
     final lastReportCount = await UserManager.getLastReportCount() ?? 0;
     final lastReviewRequest = await UserManager.getLastReviewRequest() ?? 0;
--- a/lib/utils/Utils.dart
+++ b/lib/utils/Utils.dart
@@ -11,6 +11,7 @@ import 'package:geocoding/geocoding.dart';
 import 'package:geolocator/geolocator.dart';
 import 'package:google_maps_flutter/google_maps_flutter.dart';
 import 'package:image_picker/image_picker.dart';
+import 'package:in_app_review/in_app_review.dart';
 import 'package:mosquito_alert_app/api/api.dart';
 import 'package:mosquito_alert_app/models/question.dart';
 import 'package:mosquito_alert_app/models/report.dart';
@@ -1196,5 +1197,18 @@ class Utils {
     var placemarks = await placemarkFromCoordinates(lat, lon, localeIdentifier: locale);
     if (placemarks.isEmpty) { return null; }
     return placemarks.first.locality;
+  }
+
+  static void requestInAppReview() async {
+    var myReports = await ApiSingleton().getReportsList();
+    var numReports = myReports.length;
+
+    if (numReports == 3 || numReports == 4){
+      final inAppReview = InAppReview.instance;
+
+      if (await inAppReview.isAvailable()) {
+          await inAppReview.requestReview();
+      }
+    }
   }
 }
--- a/lib/utils/Utils.dart
+++ b/lib/utils/Utils.dart
@@ -1203,6 +1203,8 @@ class Utils {
     var myReports = await ApiSingleton().getReportsList();
     var numReports = myReports.length;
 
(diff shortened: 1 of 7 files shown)


lib/utils/Utils.dart
```diff
@@ -1197,4 +1198,17 @@ class Utils {
     if (placemarks.isEmpty) { return null; }
     return placemarks.first.locality;
   }
+
+  static void requestInAppReview() async {
+    var myReports = await ApiSingleton().getReportsList();
+    var numReports = myReports.length;
+
+    if (numReports % 3 == 0 || numReports == 4){
```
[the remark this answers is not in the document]
DigitalSeneca: > This would introduce a bug, since numReports = 0 would trigger this if statement

I assumed that since this method is executed after a report is created, then the number would always be 1 or higher. But yes, it's always better to check and be sure. In this case, if I return when the list is smaller than 3, the bug is gone too. Does it make sense to use the "lastReviewRequest > 2 weeks" if the numReports is less than 3? (If someone creates reports, but then they delete them)

```
final prefs = await SharedPreferences.getInstance();
final lastReviewRequest = prefs.getInt(lastReviewRequestDatetime) ?? 0;
final lastReportCount = prefs.getInt(lastReviewRequestReportCountKey) ?? 0;
```
It would make sense to do that in UserManager.dart to keep consistency with the previous code

```
if (numReports == 3 || numReports == 4) {
  shouldRequestReview = true;
} else if (numReports >= lastReportCount + 3 || now - lastReviewRequest >= 14 * 24 * 60 * 60 * 1000) {
  shouldRequestReview = true;
}
```
Maybe it makes it more logic to put in the same `if` the 3 conditionals related to `numReports` and in the second one the one related with time, but that's not super important

lib/utils/Utils.dart
```diff
@@ -1197,4 +1198,33 @@ class Utils {
...
     return placemarks.first.locality;
   }
+
+  static void requestInAppReview() async {
+    var myReports = await ApiSingleton().getReportsList();
+    var numReports = myReports.length;
+
+    if (numReports < 3) { return; }
+
+    var shouldRequestReview = false;
```
epou, line 1208: Maybe using one-line?

```dart
var shouldRequestReview = (numReports == 3 || numReports == 4 || numReports >= lastReportCount + 3 || now - lastReviewRequest >= 14 * 24 * 60 * 60 * 1000);
```

lib/utils/Utils.dart
```diff
@@ -1197,4 +1198,33 @@ class Utils {
...
+
+    if (!shouldRequestReview) { return; }
+
+    final inAppReview = InAppReview.instance;
+
+    if (await inAppReview.isAvailable()) {
+      await inAppReview.requestReview();
+
+      await UserManager.setLastReviewRequest(now);
+      await UserManager.setLastReportCount(lastReportCount);
```
epou, line 1227: use `numReports` instead of `lastReportCount`

lib/utils/Utils.dart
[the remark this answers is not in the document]
epou: > Does it ever make sense to use the "lastReviewRequest > 2 weeks" if the numReports is less than 3? (If someone creates reports, but then they delete them)

Only once numReports is greater than 3

lib/utils/Utils.dart
DigitalSeneca: seems good to me, let's keep this one-line

PR #192, opened 2024-07-17
Home page: less colors in buttons + add map button

### LINKED ISSUE #191: Home page: less colors in buttons + add map button
The 3 action buttons should be colored only on the left part.
Add a new 'Map' button that redirects to the map view (string_id: `public_map_tab`)

MA mainscreen

DigitalSeneca:
fixes #191 

new_public_map_2


epou (commented):
Looks good to me! Only that the icon should be changed.

DigitalSeneca:
I've added the new map svg icon
image

epou (approved):

PR #193, opened 2024-07-17
Auto-tagging - Avoid double hashtag

### LINKED ISSUE #190: Auto-tagging menu behaviour
When adding a new tag:
- When pressing 'enter' it remains saved but the Auto-Tagging menu folds. It should keep unfolded.
- When pressing on 'space bar', the tag appears in a new box but it is not saved (the counter does not change and when you fold and unfold the menu again it does not appear)

Users until now (schools) have been told that they should use '#' as a prefix, and now doing so appears two hashtags. Can this be avoided?

The button with the 'X' is gray, but it should be white (always the same color as text).

By default, if there is already a tag activated, the Auto-Tagging menu should appear unfolded.

When deleting a tag:
- It is expected that tags can be deleted using the 'delete' key on the keyboard.
- Clicking the 'X' button deletes the tag but closes the menu, it should not close.

DigitalSeneca:
fixes #190 -> Users until now (schools) have been told that they should use '#' as a prefix, and now doing so appears two hashtags. Can this be avoided?

avoid_double_hashtag

### DIFF
--- a/lib/pages/settings_pages/components/hashtag.dart
+++ b/lib/pages/settings_pages/components/hashtag.dart
@@ -142,6 +142,10 @@ class _StringMultilineTagsState extends State<StringMultilineTags> {
                     ),
                     onChanged: inputFieldValues.onTagChanged,
                     onSubmitted: (String newTag){
+                      if (newTag.startsWith('#')) {
+                        newTag = newTag.substring(1);
+                      }
+
                       if (!hashtags.contains(newTag)) {
                         inputFieldValues.onTagSubmitted(newTag);
                         hashtags.add(newTag);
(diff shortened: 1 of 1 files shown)


lib/pages/settings_pages/components/hashtag.dart
```diff
@@ -142,6 +142,10 @@ class _StringMultilineTagsState extends State<StringMultilineTags> {
                     ),
                     onChanged: inputFieldValues.onTagChanged,
                     onSubmitted: (String newTag){
+                      if (newTag.startsWith('#')) {
+                        newTag = newTag.substring(1);
+                      }
```
epou, line 147: I just realized that if there's a '#' in the middle of the string, it would generate two tags in the DB.

I was thinking of removing any '#' found in the string by always running `newTag = newTag.replaceAll('#', '');`

What do you think?

DigitalSeneca: It depends on how we want to handle it. I guess it's not expected that any user would want to add '#' in the middle of a hashtag. Removing the symbols and combining it into a single hashtag seems like a good solution from the developers perspective

epou (approved):

PR #199, opened 2024-08-01
Added CI: sync tags localise

epou:


.github/workflows/sync_tags_localise.yml
```diff
@@ -0,0 +1,46 @@
+name: Synchronize Github Tags with Localise tags
+
+on:
+    workflow_call:
+    workflow_dispatch:
+    push:
+        tags:
+            - '*'
```
DigitalSeneca, line 8: The update_locales.yml is also executed when we push a tag, and it calls this workflow so this will be executed twice. In this case, we should remove those 3 lines so it's only called once

.github/workflows/update_locales.yml
```diff
@@ -0,0 +1,57 @@
...
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v4
+
+            - uses: actions/setup-python@v5
+              with:
+                python-version: '3.10'
+                # cache: 'pip'
```
DigitalSeneca, line 29: should we remove this comment?

.github/workflows/update_locales.yml
```diff
@@ -0,0 +1,57 @@
...
+
+            - name: Install dependencies
+              run: |
+                python -m pip install --upgrade pip
+                pip install requests
+                # pip install -r requirements.txt  # install your dependencies
```
DigitalSeneca, line 35: remove this commented line?

.github/workflows/update_locales.yml
```diff
@@ -0,0 +1,57 @@
...
+        needs: [sync_localise]  # require sync_localise to pass before deploy runs
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v4
+
+            - uses: actions/setup-python@v5
+              with:
+                python-version: '3.10'
```
DigitalSeneca, line 28: If there are no incompatibilities, maybe we can use python v3.11 or v3.12 so end-of-life arrives 1 or 2 years later and we can reduce maintainability efforts. Otherwise, 3.10 should also work fine
image

DigitalSeneca (commented):
Very nice! This will be helpful

.github/workflows/sync_tags_localise.yml
```diff
@@ -0,0 +1,46 @@
...
+              with:
+                  url: 'https://localise.biz/api/tags'
+                  method: 'POST'
+                  customHeaders: '{"Authorization": "Loco ${{ secrets.LOCALISE_FULL_API_KEY }}"}'
+                  contentType: "application/x-www-form-urlencoded"
+                  data: '{"name": "${{ github.ref_name }}"}'
+
+            - name: Getting all assets being used
+              run: |
+                echo "ASSET_LIST=$(grep -hrPo "MyLocalizations\.of\s*\(\s*context\s*,\s*['\"]\K[^'\"]*[^'\"]" lib/ | sort | uniq | paste -s -d,)" >> $GITHUB_ENV
```
epou, line 37: This bash command catches almost all string_ids, but some are not... If you find @DigitalSeneca a better approach let me know... Not sure if there's any flutter code that can get a list of all string_ids being used...

epou:
@DigitalSeneca I’d appreciate your thoughts on this. I’ve been considering the possibility of grouping string_ids in Localise, but the challenge is that Localise currently only supports grouping by tags. This led me to the idea of syncing GitHub tags with Localise tags, which would require us to establish a consistent tagging strategy in GitHub.

Another aspect I’m considering is automating pull requests based on specific actions—perhaps when a tag is created or when a new pull request is made, though I’m still exploring the best triggers to use.

At the moment, we’re using both GitHub tags and branches for each version, and I’m uncertain about which triggers would be most effective. I’d love to hear your perspective on this.

- When creating new tags on Localise? Only on new github tags? What about new branches? Or new pull requests?
- When creating a new pull request with translation updates? On every pull request?

DigitalSeneca:
Maybe we can discuss about this topic more in detail/depth in a call to make a decision, but:

From the perspective of the app developer, in the current setup the most logic/comfortable would be to sync on the pull request that contain translations, or to get them in the v34 / v{x} branch, but maybe it's not the easiest to implement

> At the moment, we’re using both GitHub tags and branches for each version

Maybe it's easier/simpler to change the workflow to remove the extra long-term branches (like v34) and target master so that we can synchronize Localise using Github Tags only instead of also new branches

epou:
> Maybe it's easier/simpler to change the workflow to remove the extra long-term branches (like v34) and target master so that we can synchronize Localise using Github Tags only instead of also new branches

That makes a lot of sense to me :)

update_locales.py
```diff
@@ -25,11 +25,28 @@ def update_locales(api_key: str, path: Path, format: str) -> None:
...
     session.auth = (api_key, '')
 
+    if tags:
+        # Check which of the tags passed exist
+        available_tags = session.get('https://localise.biz/api/tags').json()
+
+        # Keep only the tags found in both lists (server and input parameter)
+        tags = list(
+            set(tags).intersection(set(available_tags))
+        )
```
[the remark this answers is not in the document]
epou: I believe it's acceptable for the request to fail in this case, as it allows us to understand the reason behind the failure. Additionally, I’m not sure if handling this type of request within the CI workflow is the best approach, as repeatedly retrying could lead to unnecessary complications.

.github/workflows/update_locales.yml
epou: Solved in: a331728

.github/workflows/update_locales.yml
epou: Solved in: a331728

.github/workflows/update_locales.yml
epou: Resolved at 8fde9ed

DigitalSeneca (approved):
nice!

PR #206, opened 2024-08-16
Compress images before sending.

epou:


epou:
@DigitalSeneca , I need from you to test it. (pubspeck-lock might need a commit with changes)
 
- Aspect ratio is the same as before compressing.
- EXIF data is preserved (important for us is date and GPS).
- Some models take HEIC pictures. Ensure they are transcoded to JPEG.
- Big images are indeed compressed.

DigitalSeneca:
> (pubspeck-lock might need a commit with changes)

I've added the commit with the pubspec.lock that also includes the changes from master and from this PR
I still have to test more in detail the points you mentioned

> Some models take HEIC pictures. Ensure they are transcoded to JPEG.

Do you know some model examples so that I can test with them?

epou:
Looks like samsung can, but needs to be enabled.

epou:
Looks like samsung can, but needs to be enabled.
https://us.community.samsung.com/t5/Galaxy-S23/Pictures-switched-to-HEIC-how-can-I-get-it-back-to-JPEG/td-p/2653534

DigitalSeneca:
> Looks like samsung can, but needs to be enabled. https://us.community.samsung.com/t5/Galaxy-S23/Pictures-switched-to-HEIC-how-can-I-get-it-back-to-JPEG/td-p/2653534

I don't have that option in my physicial phone or in the emulator, and I also can't take high resolution pictures since my phone is not high-end, do you have available the picture that created the issue in production in the first place? That would be amazing since it would have the EXIF data and all the info needed. Alternatively I could get some image from google and select it from the gallery to test the aspect ratio, but there would be no EXIF data

epou:
I see them compressed when I compare the original vs the ones you've shared in the previous comments. All of them are now 4k instead, in jpeg format and has been reduced in size. The only thing I see is that in some pictures EXIF data is not there...

DigitalSeneca:
I've looked more in detail the GPS info in EXIF and I am observing that the first image (`Big image (108MP) with GPS in EXIF: https://webserver.mosquitoalert.com/media/tigapics/6a3af8ef-4e51-4fbf-bb0c-a1595dd9f618.jpg`) does not have GPS data and that seems to be the reason why it also does not have GPS data in the one I uploaded.

But the second image that I uploaded I see that it still has the GPS data
image

PR #208, opened 2024-08-19
V34

DigitalSeneca:


DigitalSeneca:
@epou Related to the comment in [#199 - Maybe it's easier/simpler to change the workflow to remove the extra long-term branches (like v34) and target master so that we can synchronize Localise using Github Tags only instead of also new branches](https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/199#issuecomment-2284305050)

Should we start this process of using master? Or what are your ideas regarding this?

epou:
@DigitalSeneca, it makes sense to me! So then, we should pay attention only to the tags regarding versioning. Merge whenever you feel comfortable

PR #217, opened 2024-11-11
v35 - Firebase - Remove subscriptions

DigitalSeneca:


lib/api/api.dart
```diff
@@ -626,35 +623,6 @@ class ApiSingleton {
     }
   }
 
-  Future<List<Topic>?> getTopicsSubscribed(String userIdentifier) async {
```
epou, line 629: so then, this should be remove too

`static const get_my_topics = '/topics_subscribed/';`

lib/api/api.dart
epou, line 1: why remove `getTopicsSubscribed` but keep `unsubscribeFromTopic` and `subscribeToTopic`?
all of them interacts with the deprecated API. Is this just an intermediate step?

PR #219, opened 2024-11-11
v35 - Upgrade flutter to stable:3.19.5 + Upgrade/Remove dependencies

DigitalSeneca:
## Description of changes

- Upgrade minor versions
- Remove unused dependencies
- Set flutter to be compatible with stable v3.19.5
- Upgrade breaking change battery_plus from 2.x.x to 6.x.x


lib/pages/settings_pages/settings_page.dart
```diff
@@ -26,30 +26,30 @@ class _SettingsPageState extends State<SettingsPage> {
...
+    Language('lb_LU', 'Luxembourgish (Luxembourg)', 'Luxembourgish (Luxembourg)'),
+    Language('mk_MK', 'Macedonian (Former Yugoslav Republic of Macedonia)', 'Macedonian (Former Yugoslav Republic of Macedonia)'),
+    Language('nl_NL', 'Dutch', 'Dutch'),
+    Language('pt_PT', 'Protuguese', 'Protuguese'),
+    Language('ro_RO', 'Romanian', 'Romanian'),
+    Language('sl_SI', 'Slovenian (Slovenia)', 'Slovenian (Slovenia)'),
+    Language('sq_AL', 'Albanian', 'Albanian'),
+    Language('sr_RS', 'Serbian', 'Serbian'),
+    Language('sv_SE', 'Swedish', 'Swedish'),
+    Language('tr_TR', 'Turkish (Turkey)', 'Turkish (Turkey)'),
```
DigitalSeneca, line 52: the new version of `language_picker` requires a new parameter `native_language`

.github/workflows/static_analysis.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     - uses: subosito/flutter-action@v2
       with:
         channel: 'stable'
-        flutter-version: '3.13.7'
```
epou, line 14: Since the decision is to start from scratch, could we use the last stable version of flutter? 3.24.0

lib/pages/settings_pages/settings_page.dart
epou: Okay, maybe we could use a specific package for managing i18n, like https://pub.dev/packages/flutter_localization or the native https://docs.flutter.dev/ui/accessibility-and-internationalization/internationalization

.github/workflows/static_analysis.yml
DigitalSeneca: It would be amazing and I tried to upgrade to 3.24.0 but it required changes with java and gradle and that's why I went with an incremental step, at least for now

epou (approved):

PR #224, opened 2024-11-29
Use TextTheme to set text styles across the app

DigitalSeneca:
| image | image |


DigitalSeneca:
@epou TextThemes is working very nice as expected. I found that removing the file style.dart creates errors in around 10 files so maybe it makes sense to have logical PR increments:
- This one that implements some of the text styles and implements it into Account Page
- New style for HomePage + directly using textTheme there
- Other pages (...)

What do you think?

epou:
Agree @DigitalSeneca ! Let's do it like this :) The idea is to eliminate the 'styles.dart' file using ThemeData instead.

epou (approved):

PR #237, opened 2025-01-17
Update .gitignore

epou:


.gitignore
```diff
@@ -15,39 +15,78 @@
...
+# Android related
+android/app/google-services.json
+**/gradle-wrapper.jar
+**/android/**/gradle-wrapper.jar
+**/android/.gradle
+**/android/captures/
+**/android/gradlew
+**/android/gradlew.bat
+**/android/key.properties
+**/android/key.jks
```
DigitalSeneca, line 45: I guess the android folder will always be at root so `android/whatever` would work

DigitalSeneca (approved):

PR #238, opened 2025-01-17
Hot fix user creation

epou:


DigitalSeneca (approved):
It was not possible to directly merge into main, so I have:
- created a tag for 1738 (hot fix)
- and solved the merge conflict (pubspec.yaml and pubspec.lock) by merging main into this branch.

This means that the branch `hot-fix-user-creation` does no longer contain changes from only v33 but I think it's acceptable because we already have the build in PlayStore (internal test)

PR #243, opened 2025-01-31
Fix UUID overflow iOS

DigitalSeneca:
image

### DIFF
--- a/lib/pages/main/drawer_and_header.dart
+++ b/lib/pages/main/drawer_and_header.dart
@@ -252,11 +252,15 @@ class _MainVCState extends State<MainVC> {
                 fontSize: 8,
               ),
             ),
-            Text(
-              snapshot.data ?? '',
-              style: TextStyle(
-                color: Colors.grey,
-                fontSize: 8,
+            Container(
+              width: 150,
+              child: Text(
+                snapshot.data ?? '',
+                overflow: TextOverflow.ellipsis,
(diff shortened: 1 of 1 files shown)


DigitalSeneca:
@epou 
Just adding an overflow to the text Widget does not work in this case because it's inside of a Row with another widget (GestureDetector) on its right, so it doesn't detect the overflow correctly. By adding a parent with a defined width, the overflow works
image

lib/pages/main/drawer_and_header.dart
```diff
@@ -252,11 +252,15 @@ class _MainVCState extends State<MainVC> {
                 fontSize: 8,
               ),
             ),
-            Text(
-              snapshot.data ?? '',
-              style: TextStyle(
-                color: Colors.grey,
-                fontSize: 8,
+            Container(
+              width: 150,
```
epou, line 256: I was wondering if the "Expanded" (or Flexible :thinking: ) class can work here. So it's not needed to force 150px width and it will adapt to different screen sizes

DigitalSeneca: I tried both but then the uuid disappears
image
We can try by using a percentage of the screen, for example 35% that look like this, what do you think:
image

epou: as you see fit

epou (approved):

PR #235, opened 2025-01-17
Merging iOS changes

epou:

PR #245, opened 2025-02-06
CI/CD: Compilation test Android

DigitalSeneca:


android/app/google-services.json
```diff
@@ -0,0 +1,157 @@
...
+            }
+          },
+          {
+            "client_id": "960892168613-8qoen49sp0f1g9qkljvhfjjsk7r18kbr.apps.googleusercontent.com",
+            "client_type": 3
+          }
+        ],
+        "api_key": [
+          {
+            "current_key": "CONFIDENTIAL-HERE-GOES-YOUR-API-KEY"
```
DigitalSeneca, line 64: According to Google (https://groups.google.com/g/firebase-talk/c/bamCgTDajkw) google-services.json is not a sensitive file so we can add it to git since it necessary to compile the app.
The only sensitive data is the api_key that appears twice on lines 64 and 133

android/app/build.gradle
```diff
@@ -34,7 +34,7 @@ if (keystorePropertiesFile.exists()) {
 
 Properties properties = new Properties()
 properties.load(project.rootProject.file('local.properties').newDataInputStream())
-def googlemapsKey = properties.getProperty('googlemaps.Key')
+def googlemapsKey = "properties.getProperty('googlemaps.Key')"
```
DigitalSeneca, line 37: maybe using github secrets or .env file

.github/workflows/build_app.yml
```diff
@@ -0,0 +1,36 @@
...
+        flutter-version: '3.27.3'
+  
+    - run: flutter pub get
+
+    - name: Build appbundle
+      run: flutter build appbundle --debug --verbose
+      #env:
+      #  GOOGLEMAPS_KEY: ${{ secrets.GOOGLEMAPS_KEY }}
+
+  ios:
```
DigitalSeneca, line 23: The iOS compilation test was successful using the code from PR #235 -> https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/actions/runs/13203112097

pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: A new Flutter project.
 version: 2.3.0+1738
 
 environment:
-  sdk: '>=2.17.1 <3.0.0'
+  sdk: '>=2.19.0 <3.0.0'
```
DigitalSeneca, line 6: It is mentioned that null safety issues could be caused by sdk <= 2.19.0
flutter 3.29.3 uses a higher sdk than that so it does not have any con

.github/workflows/build_app.yml
```diff
@@ -0,0 +1,38 @@
...
+    - uses: subosito/flutter-action@v2
+      with:
+        channel: 'stable'
+        flutter-version: '3.27.3'
+  
+    - run: flutter pub get
+
+    - name: Build appbundle (with decoded google-services.json)
+      env:
+        GOOGLE_SERVICES_JSON: ${{ secrets.GOOGLE_SERVICES_JSON }}
```
epou, line 20: secrets.GOOGLE_SERVICES_JSON_ANDROID_BASE64

analysis_options.yaml
epou, line 1: Is this the only file "using" pedantic? So then, after the removal, 'pedantic' can also be removed from pubspec.yaml

DigitalSeneca: I'd say we are not using pedantic and it's actually a discountinued package, so removing it completely seems a good idea

epou (approved):

PR #249, opened 2025-02-12
Fix Flutter analyze not detecting "warning" and "error" due to formatting

DigitalSeneca:


.github/workflows/analyze_and_test.yml
```diff
@@ -15,10 +15,15 @@ jobs:
 
     - name: Statically analyze the Dart code for any errors/warnings, ignoring "info" level warnings
       run: |
-        if flutter analyze 2>&1 | grep -q -E 'error:|warning:'; 
-        then exit 1
-        else exit 0
+        if flutter analyze 2>&1 | grep -q -E 'error|warning';
```
epou, line 18: Instead of running "flutter analyze" twice (case when error or warning is found) to show the results in the github actions logs, I suggest doing something like:
```bash
output=$(flutter analyze 2>&1)
echo "$output"  # Always print the analysis output

if echo "$output" | grep -q -E 'error|warning'; then
  exit 1
else
  exit 0
fi
```

epou: Ah! Can you try with:
```bash
    flutter analyze 2>&1 | tee flutter_analyze.log
    if grep -q -E -i 'error|warning' flutter_analyze.log; then
        exit 1
    else
        echo "No errors or warnings found."
        exit 0
    fi
```

epou, line 18: add `-i` in grep to ignore case

epou (approved):

PR #251, opened 2025-02-13
iOS changes (compile + background tracking)

DigitalSeneca:


ios/Podfile.lock
```diff
@@ -0,0 +1,354 @@
...
+    - libwebp (~> 1.0)
+    - SDWebImage/Core (~> 5.17)
+  - shared_preferences_foundation (0.0.1):
+    - Flutter
+    - FlutterMacOS
+  - sqflite (0.0.3):
+    - Flutter
+    - FlutterMacOS
+  - SwiftyGif (5.4.4)
+  - url_launcher_ios (0.0.1):
```
DigitalSeneca, line 192: this file needs to be regenerated by a developer with a Mac because we are not using this dependency

PR #252, opened 2025-02-13
Changes to compile iOS + permissions for Background tracking on iOS

DigitalSeneca:


lib/main.dart
```diff
@@ -1,10 +1,12 @@
 import 'dart:async';
+import 'dart:io';
```
epou, line 2: I'm not sure about this import

lib/pages/settings_pages/settings_page.dart
```diff
@@ -1,3 +1,5 @@
+import 'dart:io';
```
epou, line 1: I'm not sure about this import

DigitalSeneca: This one is needed to call "Platform.isAndroid" - "Platform.isIOS"

lib/main.dart
DigitalSeneca: I've removed it

PR #253, opened 2025-02-18
flutter analyze - Fix warnings

DigitalSeneca:


lib/api/api.dart
```diff
@@ -68,8 +67,6 @@ class ApiSingleton {
     'Authorization': 'Token ' + token
   };
 
-  final FirebaseAuth _auth = FirebaseAuth.instance;
```
DigitalSeneca, line 71: warning - The value of the field '_auth' isn't used - lib\api\api.dart:71:22 - unused_field

lib/api/api.dart
```diff
@@ -652,7 +649,7 @@ class ApiSingleton {
             'Request: ${response.request.toString()} -> Response: ${response.body}');
         return ApiResponse.fromJson(json.decode(response.body));
       }
-      Map<String, dynamic>? jsonAnswer = json.decode(response.body);
+      json.decode(response.body);
```
DigitalSeneca, line 652: warning - The value of the local variable 'jsonAnswer' isn't used - lib\api\api.dart:655:29 - unused_local_variable

lib/pages/settings_pages/campaign_tutorial_page.dart
```diff
@@ -4,7 +4,7 @@ import 'package:mosquito_alert_app/utils/MyLocalizations.dart';
 import 'package:mosquito_alert_app/utils/style.dart';
 
 class CampaignTutorialPage extends StatefulWidget {
-  bool? fromReport;
+  final bool? fromReport;
```
DigitalSeneca, line 7: warning - This class (or a class that this class inherits from) is marked as '@immutable', but one or more of its instance fields aren't final: CampaignTutorialPage.fromReport -
       lib\pages\settings_pages\campaign_tutorial_page.dart:6:7 - must_be_immutable

lib/utils/PushNotificationsManager.dart
```diff
@@ -159,7 +159,7 @@ class PushNotificationsManager {
 
   static Future<void> registerFCMToken(String fcmToken) async {
     var userId = await UserManager.getUUID();
-    var result = ApiSingleton().setFirebaseToken(userId, fcmToken);
+    ApiSingleton().setFirebaseToken(userId, fcmToken);
```
DigitalSeneca, line 162: warning - The value of the local variable 'result' isn't used - lib\utils\PushNotificationsManager.dart:158:9 - unused_local_variable

lib/utils/Utils.dart
```diff
@@ -965,26 +965,24 @@ class Utils {
   }
 
   static ui.Locale getLanguage() {
-    if (ui.window.locale != null) {
```
DigitalSeneca, line 968: warning - The operand can't be 'null', so the condition is always 'true' - lib\utils\Utils.dart:968:26 - unnecessary_null_comparison

lib/api/api.dart
epou: Can we remove this piece of code? `json.decode(response.body)` will do nothing aside from serializing the response, but it won't be used anywhere

DigitalSeneca: I think removing this line would alter the functionality because if json.decode fails, it would go to "catch" and return false instead of true

PR #254, opened 2025-02-18
Run CI on every commit to main branch

epou:


.github/workflows/build_app.yml
```diff
@@ -1,6 +1,8 @@
 name: Build app
 on:
   pull_request:
+  push:
+    branches: ["main"]
```
DigitalSeneca, line 5: would it make sense to enable a merge_queue for the main branch so that if the workflow fails, the PR is not merged?

epou: 'main' branch already 'Require status checks to pass before merging'. I've never worked with merging queues before... As far as I see in the [documentation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request-with-a-merge-queue):
> Using a merge queue is particularly useful on branches that have a relatively high number of pull requests merging each day from many different users.

...which might be not our case.


If you think it can be beneficial, please let me know and we can adapt to start using merging queues

DigitalSeneca: True, that's not our case. Another case that may be beneficial is if we have longer integration tests, but not our case at the moment.
If we have 'Require status checks to pass before merging', then that looks good to me!

DigitalSeneca (approved):

PR #256, opened 2025-02-19
Reduced size of large images

epou:
```bash
for img in assets/img/sendmodule/*.png; do
  convert "$img" -resize 800x936\> -quality 60 "${img%.png}_resized.png"
  mv "${img%.png}_resized.png" "$img"
done
```


DigitalSeneca (approved):
I checked how the reduced images look in the emulator and looks good!

PR #257, opened 2025-02-19
Removed unused tutorial

epou:


DigitalSeneca (approved):
nice, the app size will go down in size a few megabytes with the last changes

PR #266, opened 2025-02-25
Use Flutter's SDK version settings in build.gradle

epou:
See: https://stackoverflow.com/questions/76108428/how-do-i-fix-namespace-not-specified-error-in-android-studio

### DIFF
--- a/android/settings.gradle
+++ b/android/settings.gradle
@@ -18,7 +18,7 @@ pluginManagement {
 
 plugins {
     id "dev.flutter.flutter-plugin-loader" version "1.0.0"
-    id "com.android.application" version '7.4.2' apply false
+    id "com.android.application" version '8.1.4' apply false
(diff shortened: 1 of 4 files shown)


DigitalSeneca (changes requested):
The CI job "build app android" failed, so it means this change is not using the same sdk number

android/settings.gradle
```diff
@@ -18,7 +18,7 @@ pluginManagement {
 
 plugins {
     id "dev.flutter.flutter-plugin-loader" version "1.0.0"
-    id "com.android.application" version '7.4.2' apply false
+    id "com.android.application" version '8.1.4' apply false
```
epou, line 21: Current flutter targetSdkVersion is 35, which requires AGP version 8.1.0 or greater.

DigitalSeneca (approved):
I checked it and is working fine for me

PR #267, opened 2025-02-25
Using version and build number from pubspec.yml

epou:


epou:
@DigitalSeneca , should I update the 'version' in pubspec.yml from 2.4.0+1740 to 4.1.0+2872 (according to #265 )

epou:
@JohnPalmer @StuffJoy , could you try if this works OK for Xcode/ios?

PR #268, opened 2025-02-27
Upgrade gradle to 7.4.2

epou:


android/app/build.gradle
```diff
@@ -48,15 +49,10 @@ android {
...
-    lintOptions {
-        disable 'InvalidPackage'
-        checkReleaseBuilds false
-        abortOnError false
-    }
 
     defaultConfig {
         applicationId "ceab.movelab.tigatrapp"
-        minSdkVersion flutter.minSdkVersion
+        minSdkVersion 24
```
epou, line 55: Lower version break the build

DigitalSeneca: In a previous PR we updated the code to not have the number hardcoded and use flutter.minSdkVersion instead. Is there not a possibility to continue using it and update to 24 somewhere else?

DigitalSeneca (commented):
I tested it locally and no issues from my part

android/app/build.gradle
epou: This is the error I was having if using a min SDK lower than 24

https://github.com/flutter/flutter/issues/146038

I use a suggestion in a comment to use minsdk 24 and it solved the problem. Do you have any idea?

epou: Is 21 for flutter 3.27.3
https://github.com/flutter/flutter/blob/c519ee916eaeb88923e67befb89c0f1dabfa83e6/packages/flutter_tools/gradle/src/main/groovy/flutter.groovy#L48

DigitalSeneca: I checked that for flutter 3.29 (latest version) the minSdk is also 21, so yeah, hardcoding it to 24 would be the solution in this case.
Just consider that after this change, devices from years 2014-2015 will be incompatible (less than 3% of total userbase)
https://apilevels.com/

DigitalSeneca (approved):

PR #270, opened 2025-02-27
Removed firebase_auth

epou:


ios/Podfile.lock
DigitalSeneca, line 1: This file is created automatically by the command "pod install" on macos, isn't it? If it's like pubspec.lock, we should not edit manually

epou: We can only do that from a mac device... Should I restore the changes on that files then?

DigitalSeneca: pff yeah I know it does not depend on either you or me, it should be someone with a mac device.
I would think it's likely to work when removing it manually and it's better to do it manually than not doing it

DigitalSeneca (approved):

PR #274, opened 2025-02-27
Update minimum dart version to 3.2

epou:


pubspec.lock
DigitalSeneca, line 1: When I run it locally I get a different pubspec.lock 🤔

pubspec.yaml
```diff
@@ -5,7 +5,7 @@ publish_to: none
 
 
 environment:
-  sdk: '>=2.19.0 <3.0.0'
+  sdk: '>=3.2.0 <4.0.0'
```
DigitalSeneca, line 8: For reference, I am using flutter 3.27.4 - dart 3.6.2
https://docs.flutter.dev/release/archive

epou: I was using a different flutter version at that time. That's why I opened https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/275

epou:
Closed in favor of #275 , which already upgrade the dart code

PR #275, opened 2025-02-27
Start using FVM

epou:


ios/Podfile.lock
DigitalSeneca, line 1: we don't need this file anymore?

epou: This file is autogenerated during the `flutter build ios`. If we keep it in the repository, any changes to pubspec.yaml will make CI (build_ios) break as we are not able to run `pod install` locally and commit the changes.

DigitalSeneca (commented):
Can you update README.md line 20 from "flutter run" to "fvm flutter run"?
Maybe more lines need to be adjusted to reflect fvm

.fvmrc
```diff
@@ -0,0 +1,3 @@
+{
+  "flutter": "3.27.3"
```
DigitalSeneca, line 2: I only see the version defined, but not the branch. Is it implied that it's always branch "stable"?

I tried running "fvm use 3.27.3@stable" and got:

´´´
{
  "flutter": "3.27.3@stable"
}
´´´
I got the same in fmv_config.json

ios/Podfile.lock
DigitalSeneca: understood

.fvmrc
DigitalSeneca: Ah perfect, then it's good to go like that. I would still prefer to mark it explicitly just in case but that's subjective

DigitalSeneca (approved):

.fvmrc
epou: Sure! I'm doing it right now

epou: See: https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/275/commits/8fa72203f18dd0e92bf9d137fc7c90b6517893f8

PR #296, opened 2025-03-05
Changed tab icons

epou:


DigitalSeneca (commented):
for frontend changes could you add screenshots like I usually do? This way would be easier to check

I run it locally and it seems that the Tile background color has the same problem with color we saw in previous days?
image

Apart from that and that we're waiting for the voting on the tab colors, it looks great

DigitalSeneca (approved):

PR #297, opened 2025-03-05
Added GoogleService-Info.plist to gitignore

epou:


.gitignore
```diff
@@ -90,6 +90,7 @@ android/JRBP_Keystore
 
 # Service account files
 svc-keyfile.json
+GoogleService-Info.plist
```
DigitalSeneca, line 93: We already have this file on line 82. Also on line 82 it's at the expected iOS location but here it would check for it at the directory root

PR #300, opened 2025-03-05
'PREV' button using same style + replace deprecated MaterialStateProperty

DigitalSeneca:
image


epou (approved):
My guess here is that you're removing renderSkipBtn() since they are not used

DigitalSeneca:
> My guess here is that you're removing renderSkipBtn() since they are not used

that's correct

PR #302, opened 2025-03-06
Added firebase_analytics

epou:
Enable debug:
`adb shell setprop debug.firebase.analytics.app ceab.movelab.tigatrapp`
Disable debug:
`adb shell setprop debug.firebase.analytics.app .none.`


lib/main.dart
```diff
@@ -81,6 +82,15 @@ class _MyAppState extends State<MyApp> {
 
   MyLocalizationsDelegate _newLocaleDelegate = MyLocalizationsDelegate();
 
+  static FirebaseAnalytics analytics = FirebaseAnalytics.instance;
+  static FirebaseAnalyticsObserver observer =
+      FirebaseAnalyticsObserver(
+        analytics: analytics,
```
DigitalSeneca, line 88: if it's only used once, do we really need to create a variable?

```
  static FirebaseAnalyticsObserver observer =
      FirebaseAnalyticsObserver(
        analytics: FirebaseAnalytics.instance,
```

DigitalSeneca (commented):
From a technical point of view, this is working fine for me too

From a legal point of view, is sending information for analysis already covered in the current agreement that the users are accepting? This is outside my area of expertise but many apps asks this when signing up, so it's worth to consider if it's needed

lib/main.dart
epou: I'll change it :+1:

PR #303, opened 2025-03-06
Changed icon and position for public map button

epou:
Screenshot_1741269584


DigitalSeneca (approved):
I like it, but I thought the feedback was to put the map more down compared to the other buttons

DigitalSeneca (approved):
I like it, but I thought the feedback was to put the map button more down compared to the other buttons

epou:
Yes I know... But the team has decided to place the map button along with the others. Without any separation...

PR #319, opened 2025-03-17
My Reports: Remove purplish background color

DigitalSeneca:


pubspec.lock
epou, line 1: have you used FVM? i see package downgrading to lower versions

DigitalSeneca: Running "fvm flutter pub get" generate the file. This pull request does not modify dependency versions, so you're right in this sense that it should not downgrade versions. So I have gone to main, execute "fvm flutter pub get" there again and still produces this modified pubspec.lock, which makes me think that a previous PR didn't update this file

epou (approved):

PR #323, opened 2025-03-20
Move all logic to BackgrounTracking class + start early tracking

epou:


lib/utils/BackgroundTracking.dart
```diff
@@ -1,60 +1,265 @@
...
+      }
+
+      if (alwaysStatus.isPermanentlyDenied) {
+        await openAppSettings();
+      }
+    } else {
+      print('Location When In Use permission denied. Tracking not enabled.');
+    }
+
+    return Permission.locationAlways.status.isGranted;
```
DigitalSeneca, line 108: is this actually true? It seems like it's always returning the same value even when the user rejects the permission

lib/utils/BackgroundTracking.dart
```diff
@@ -1,60 +1,265 @@
...
+      await prefs.setString('trackingUUID', trackingUuid);
+    }
+    return trackingUuid;
+  }
+
+  static Future<String> getSchedulerTaskQueue() async {
+    List<String> test = await BackgroundTracking._getSchedulerTaskQueue();
+    bool test1 = await BackgroundTracking.canScheduleAt(DateTime.now());
+    String myId = await BackgroundTracking._getSchedulerId(DateTime.now());
+    return test.join(',') + test1.toString() + myId;
```
DigitalSeneca, line 229: before merging, rename test to a meaningful name

epou: ups, this function was for testing. will remove...

lib/utils/BackgroundTracking.dart
epou: I've made some change into the logic in commit 009d67. Can you check again?

lib/utils/BackgroundTracking.dart
```diff
@@ -1,60 +1,258 @@
...
+
+    BackgroundTracking._clearSchedulerTaskQueue();
+  }
+
+  static Future<bool> isEnabled() async {
+    var prefs = await SharedPreferences.getInstance();
+    return prefs.getBool(BackgroundTracking._isEnabledPrefsKey) ?? false;
+  }
+
+  static Future<bool> _hasLocationPermissions() async {
```
DigitalSeneca, line 80: The method _hasLocationPermissions could be renamed to _hasLocationAlwaysPermission because it specifically checks for the locationAlways permission.
```
static Future<bool> _hasLocationAlwaysPermission() async {
```

lib/utils/BackgroundTracking.dart
```diff
@@ -1,60 +1,258 @@
...
+      await openAppSettings(); // Open app settings for the user to manually enable the permission
+    }
+
+    if (await Permission.locationWhenInUse.request().isGranted) {
+      // The locationAlways permission can not be requested directly, the user
+      // has to request the locationWhenInUse permission first. Accepting this
+      // permission by clicking on the 'Allow While Using App' gives the user
+      // the possibility to request the locationAlways permission.
+      if (!await Permission.locationAlways.request().isGranted) {
+        await openAppSettings();
```
DigitalSeneca, line 102: If the user has given permission to "locationWhenInUse" (lines 93 and 96), then here we should not need to openAppSettings, it would be more user friendly to just request `Permission.locationAlways.request()`

lib/utils/BackgroundTracking.dart
epou: Done

lib/utils/BackgroundTracking.dart
epou: This is what is done here. It tries to request locationAlways, and if not granted it opens the app settings. `request()` does not prompt the user if that permission is already given.
Excerpt from the docu:
```
Call request() on a Permission to request it. If it has already been granted before, nothing happens.
request() returns the new status of the Permission.

if (await Permission.contacts.request().isGranted) {
  // Either the permission was already granted before or the user just granted it.
}
```

Let me know if I have answered your comment

DigitalSeneca: oh you are right. I misread line 101 like if it was `if (!await Permission.locationAlways.isGranted) {`

lib/pages/settings_pages/settings_page.dart
```diff
@@ -144,21 +143,32 @@ class _SettingsPageState extends State<SettingsPage> {
...
+                              ? CircularProgressIndicator()
+                              : null,
                           onChanged: (bool value) async {
-                            await UserManager.setTracking(value);
-                            var trackingStatus =
-                                await UserManager.getTracking();
+                            if (value) {
+                              setState(() {
+                                isBgTrackingLoading = true;
+                                isBgTrackingEnabled = true;
```
DigitalSeneca, line 153: duplicated line, remove 1

lib/pages/settings_pages/settings_page.dart
```diff
@@ -144,21 +143,32 @@ class _SettingsPageState extends State<SettingsPage> {
...
+                              });
+                              await BackgroundTracking.start(shouldRun: true)
+                                  .whenComplete(() {
+                                setState(() {
+                                  isBgTrackingLoading = false;
+                                });
+                              });
+                            } else {
+                              setState(() {
+                                isBgTrackingEnabled = false;
```
DigitalSeneca, line 163: I guess this is redundant since lines 167-171 will already set the state to false

lib/utils/BackgroundTracking.dart
```diff
@@ -1,60 +1,277 @@
...
+        await Workmanager().registerOneOffTask(
+          'tracking_task_${scheduledTime.millisecondsSinceEpoch}',
+          'trackingTask',
+          initialDelay: scheduledTime.difference(now),
+          tag: 'trackingTask',
+          constraints: Constraints(networkType: NetworkType.connected),
+        );
+      });
+
+      return true;
```
DigitalSeneca, line 204: should this be a Future<true>?

lib/pages/settings_pages/settings_page.dart
epou: It's not duplicated, one is for loading (show spinner) and the other for enabled (toggle value)

lib/utils/BackgroundTracking.dart
epou: I was thinking that when using `async` for a function it implicitly returns a Future and Dart automatically wraps it in a Future since the function is async. Is that true?

lib/pages/settings_pages/settings_page.dart
epou: The idea behind this is to immediately switch the toggle button, and if something goes wrong (i.e., if BackgroundTracking.isEnabled() doesn't reflect the expected value), the toggle will revert back to its original state. This approach helps users understand that an attempt was made to change the state, but an issue occurred, and the state has been restored to its previous value.

DigitalSeneca: good idea!

PR #324, opened 2025-03-20
Ask about location on LocationForm only

epou:


lib/pages/forms_pages/components/biting_location_form.dart
DigitalSeneca, line 1: wouldn't it be required for all forms? (adult, site, bite)

epou: all of the uses the class BitingLocationForm when asking for the GPS. It's quite confusing that the GPS form page is called BitingLocationForm

DigitalSeneca: ah okey, yeah it's confusing

DigitalSeneca (approved):
I want to clarify, since the requirements are that background tracking is enabled from start but it would not be possible to get locations without this permission enabled from start.

Is the expected behavior that background tracking will be disabled until the user makes their first report?

epou:
Location always is still being requested here on the first start (#323)

https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/323/files#diff-595a9d1086c0fa3ebf610da8831796697d83b59281d95ccdb4ff5f068a1e1796

The idea here is to split the request for locationAlways needed for background tracking (request on first start) from the locationWhileInUse required (optional) for reports, which can be asked later (on the location form shown), so it's easier for the user to understand why the location is being asked at that moment.

PR #326, opened 2025-03-21
Rename BitingLocationForm to GPSForm

DigitalSeneca:


epou:
I think using LocationForm instead GPSForm is a better choice. The flutter packages refer always to location and never to GPS, which can be confusing. 

Also GPS sound like using strickly satellite information, while geolocator can also access other sources (galileo, glonass, etc). Besides, GPS is the name for a geoposition service developed in the US.

DigitalSeneca:
@epou I have renamed it, and also the analytics tags from 'report_add_gps' to 'report_add_location'

PR #338, opened 2025-03-28
Performance issue on HomePage when creating a new report of any type

### LINKED ISSUE #336: Performance issue on HomePage when creating a new report of any type
**Describe the bug**

**To Reproduce**
1. Go to HomePage
2. Click on creating a new report (mosquito, bite or site) 
3. See error: It takes a few seconds to load

**Expected behavior**
It should load almost instantly

**Screenshots**
Image

**Smartphone (please complete the following information):**
Happened on both Android emulator and Android physical device

DigitalSeneca:
fixes #336 

The performance dropped significantly after this commit: https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/324/files
When debugging, I see that the bottleneck is specifically at Utils.createNewReport because those changes are checking the location every time, which is an expensive operation and results in taking multiple seconds to start a report.

| Main branch  | This pull request |
| Image | performance-issue-after  |

### DIFF
--- a/lib/pages/main/drawer_and_header.dart
+++ b/lib/pages/main/drawer_and_header.dart
@@ -88,6 +88,7 @@ class _MainVCState extends State<MainVC> {
     UserManager.userScore = await ApiSingleton().getUserScores();
     await UserManager.setUserScores(UserManager.userScore);
     await Utils.loadFirebase();
+    await Utils.getLocation(context);
     return true;
   }
 
--- a/lib/pages/main/drawer_and_header.dart
+++ b/lib/pages/main/drawer_and_header.dart
@@ -88,7 +88,6 @@ class _MainVCState extends State<MainVC> {
     UserManager.userScore = await ApiSingleton().getUserScores();
     await UserManager.setUserScores(UserManager.userScore);
     await Utils.loadFirebase();
-    await Utils.getLocation(context);
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


epou:
Good catch! The goal of #336 was to avoid requesting location permissions at startup. Instead, it was moved to the LocationForm, making it clearer to users why/when the permission is needed. Requesting location access on the home page (where no action is being taken) could confuse users and make them deny the permission.

For this PR, I would rather remove the Utils.getLocation (thus not asking for perms at init) while keeping the code removal from Utils.createNewReport, as that seems to be the part causing the issue when clicking the buttons.

DigitalSeneca:
> as that seems to be the part causing the issue when clicking the buttons.

You're correct

> For this PR, I would rather remove the Utils.getLocation (thus not asking for perms at init) while keeping the code removal from Utils.createNewReport, 

Okey, I've done that

epou:
Nice! Has the performance improved by only removing the code at Utils.createNewReport?

DigitalSeneca:
> Nice! Has the performance improved by only removing the code at Utils.createNewReport?

Yes, that's correct, the bottleneck was there.
Re added the Utils.getLocation() was not related to the performance issue, so I've removed that part to keep functionality as expected

epou (approved):

PR #339, opened 2025-03-28
Exception has occurred. _TypeError (type 'ApiResponse' is not a subtype of type 'int?')

DigitalSeneca:


epou:
Why not change the `createSession` to return `Future<int?>`?

```dart
Future<int?> createSession(Session session) async {
    try {
      final response = await http
          .post(Uri.parse('$serverUrl$sessions'),
              headers: headers,
              body: json.encode(
                session.toJson(),
              ))
          .timeout(Duration(seconds: _timeoutTimerInSeconds));
           if (response.statusCode == 201) {
            var body = json.decode(response.body);
            return body['id'] as int;
          } else {
            print("Error: Unexpected response code ${response.statusCode}");
            print("Response body: ${response.body}");
          }
    } on TimeoutException {
      print("Error: Request timed out.");
    } on http.ClientException catch (e) {
      print("HTTP Client Exception: $e");
    } catch (e) {
      print("Error : ${e}");
    }
    return null;
  }
```

So then the `Utils.createNewReports` is simplified into:

```dart
Session session = Session(
          session_ID: sessionId,
          user: userUUID,
          session_start_time: DateTime.now().toUtc().toIso8601String());

print(language);
session.id = await ApiSingleton().createSession(session);
```

What do you think?

lib/utils/Utils.dart
```diff
@@ -92,13 +92,13 @@ class Utils {
 
       int? sessionId = response + 1;
 
-      session = Session(
+      Session session = Session(
```
epou, line 95: already defined on top

DigitalSeneca: sorry I hadn't pushed my last commit

epou (approved):

PR #353, opened 2025-04-29
Ios release 4.1.0

epou:


.gitignore
DigitalSeneca, line 1: we could leave it as it was for clean diff purposes

ios/Runner/Info.plist
epou, line 1: Not needed...

.fvmrc
DigitalSeneca, line 1: we could reset this (git restore) to keep a clean diff

DigitalSeneca:
fixed by #356

PR #356, opened 2025-05-02
iOS: Remove unneeded permissions + add iOS README instructions

DigitalSeneca:
Replace #353


pubspec.lock
epou, line 1: @DigitalSeneca , I see changes on pubspeck.lock but not on pubspec.yml. Is that something desired?

DigitalSeneca: Yes, from what I found that seems to be expected in this case because I made iOS changes that can re-resolve dependencies and update pubspec.lock

```
cd ios
pod install
```

PR #359, opened 2025-05-13
Consent screen to enable/disable background tracking on start

DigitalSeneca:
May-15-2025 13-21-28

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -11,6 +11,9 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.CAMERA" />
 
+    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES"/>
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" android:maxSdkVersion="32"/>
+
     <!-- Required for Google Analytics to access the Advertising ID on Android 13+ -->
     <uses-permission android:name="com.google.android.gms.permission.AD_ID" />
 
--- a/ios/Runner/Info.plist
+++ b/ios/Runner/Info.plist
@@ -71,23 +71,9 @@
 					<string>com.googleusercontent.apps.960892168613-7b7r7duu9j9setj1vaa2tocola7l6g44</string>
 				</array>
 			</dict>
-			<dict>
-				<key>CFBundleTypeRole</key>
-				<string>Editor</string>
-				<key>CFBundleURLName</key>
-				<string>twitter</string>
-				<key>CFBundleURLSchemes</key>
-				<array>
-					<string>twitterkit-555555555</string>
-				</array>
-			</dict>
 		</array>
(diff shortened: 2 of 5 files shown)


lib/utils/BackgroundTracking.dart
```diff
@@ -275,3 +276,36 @@ class BackgroundTracking {
     return date.millisecondsSinceEpoch.toString();
   }
 }
+
+@pragma('vm:entry-point') // Mandatory if the App is using Flutter 3.1+
+void callbackDispatcher() {
```
epou, line 281: As far as I remember, this should be defined on the main.dart file, so if ever we have other background tasks (not only from the location tracking), will need to be defined how to handle them inside this function

lib/main.dart
```diff
@@ -33,49 +31,9 @@ Future<void> main({String env = 'prod'}) async {
 
   await Workmanager().initialize(callbackDispatcher, isInDebugMode: false);
 
-  bool trackingEnabled = await BackgroundTracking.isEnabled();
```
epou, line 36: This block of code ensures that background tracking behaves correctly when the app is reopened. Specifically, it handles the case where tracking was enabled, but the user closed the app and revoked location permissions while it was not running. When the app is launched again, this logic prevents background tracking from restarting improperly and avoids any resulting exceptions.

lib/pages/settings_pages/location_consent_screen/permissions_explanation.dart
```diff
@@ -0,0 +1,83 @@
...
+Future<void> _onEnableTracking(BuildContext context) async {
+  // Show loading dialog
+  showDialog(
+    context: context,
+    barrierDismissible: false,
+    builder: (context) => const Center(child: CircularProgressIndicator()),
+  );
+
+  // Start background tracking
+  await BackgroundTracking.start(
```
epou, line 73: Have you tried splitting the code and making independent of the number of previous pages?:
```dart
 try {
    await BackgroundTracking.start(
      shouldRun: true,
      requestPermissions: true,
    );
  } finally {
    if (context.mounted) {
      Navigator.of(context, rootNavigator: true).pop(); // Ensure loading dialog is closed  always
    }
  }
  
  if (context.mounted) {
    // Return to home view (first in stack)
    Navigator.popUntil(context, (route) => route.isFirst);
  }
```

lib/main.dart
DigitalSeneca: thanks for the explanation, I have added it again

lib/utils/BackgroundTracking.dart
DigitalSeneca: Okey, makes sense. I moved it because at the moment there's only background tracking tasks so it would make sense to follow the single responsibility principle by keeping all the background tracking code in the same file

lib/pages/settings_pages/location_consent_screen/permissions_explanation.dart
DigitalSeneca: you're right. I have refactored

epou (approved):

PR #360, opened 2025-05-16
Remove iOS permissions not needed from Info.plist

DigitalSeneca:
| Before | After |
| image | image |


DigitalSeneca:
> It's not clear to me if the permission has been removed or not, as I only see that the "NS...Descriptions" have been removed. I don't own a mac station, so I can not test either.

To remove a permission we need to remove it from Info.plist, disable in Podfile, and also make sure there is no code in app invoking that permission. Things like "Apple Music" and "Contacts" are not currently used in the app but they are registered in Info.plist.
I have tested it in the iPhone.

epou (approved):

Issue #362, opened 2025-05-19
iOS camera crash


DigitalSeneca:
**To Reproduce**
1. Go to report (camera opens automatically)
2. See crash

**Screenshots**
[video available in slack - May 19 on #proj-mobile-app - Fails to upload to github]

**Smartphone (please complete the following information):**
 - Device: iPhone 11
 - OS: iOS 18
 - Version main on May 19, 2025

Issue #352, opened 2025-04-11
iOS is sending black pictures on sending reports


epou:
**Screenshots**

![Image](https://github.com/user-attachments/assets/0cc38b8a-afdd-421b-9018-042ff5c4ff95)

**Smartphone (please complete the following information):**
 - Device: iPhone 11
 - OS: iOS 18.1.1
 - Version 4.1.0

**How to reproduce**
1. Click Report mosquito
2. Open camera (Looks good)
3. Take picture -> See error here - It appears a black screen and it's sent to the server as a black screenshot

PR #364, opened 2025-05-19
iOS camera: Remove offending code + Remove code

DigitalSeneca:
## What has changed
- iOS: Fixed gesture detector: Photo is only taken when clicking the 'take pic' button on the half bottom part -> Now it makes the photo when clicking at any part of the button. This caused the bug described on #352 because there was an invisible clickable that opened the black screen instead of taking a photo
- Android & iOS: Remove not-working code related to widget that should show gallery image but shows a black screen (as seen in #352) -> Instead, gallery icon shows the phone's default gallery, or if the user has multiple galleries let's the user choose one 
- Try to remove all code related to showing latest gallery images in WhatsApp style -> so that I can rebuild the widget from scratch. The bug described in #362 is not reproducible after this 

Screenshot 2025-05-19 at 15 17 55


lib/pages/forms_pages/components/whatsapp_camera/camera_whatsapp.dart
```diff
@@ -28,54 +25,6 @@ class _WhatsAppCameraController extends ChangeNotifier {
...
-    });
-  }
-
-  Future<void> getPhotosToGallery() async {
-    final permission = await handlerPermissions();
-    if (permission) {
-      final albums = await PhotoGallery.listAlbums(
-        mediumType: MediumType.image,
-      );
-      final res = await Future.wait(albums.map((e) => e.listMedia()));
```
DigitalSeneca, line 63: this line causes the crash described in #362

epou (approved):
LGTM!

PR #370, opened 2025-05-26
last changes made before compiling build 2883 for ios

JohnPalmer:
Tnese are the changes I made for compiling ios v4.1.0 build 2883, which is currently on testflight.


ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -559,8 +561,8 @@
 				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
-				DEVELOPMENT_TEAM = 45253FTRX4;
+				CURRENT_PROJECT_VERSION = 2883;
```
epou, line 564: This was set to `$(FLUTTER_BUILD_NUMBER)` to get the build number from pubspec.yaml and sync android and ios build numbers. See https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/267

ios/Runner/Info.plist
```diff
@@ -71,9 +71,23 @@
...
+				<string>twitter</string>
+				<key>CFBundleURLSchemes</key>
+				<array>
+					<string>twitterkit-555555555</string>
+				</array>
+			</dict>
 		</array>
 		<key>CFBundleVersion</key>
 		<string>$(CURRENT_PROJECT_VERSION)</string>
+		<key>FacebookAppID</key>
```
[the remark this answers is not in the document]
JohnPalmer: Same as with Twitter - if we think we can take it out without it breaking, let's do it.

ios/Runner.xcodeproj/project.pbxproj
JohnPalmer: Sorry about that - I didn't see that. I think this ends up getting updated without me thinking when I open XCode as part of the final checks in the build process. Let's set it back the way it was to stay in sync with Android.

PR #381, opened 2025-06-11
Get user score using new api sdk + replace SharedPrefs by Provider

DigitalSeneca:


lib/providers/user_score_provider.dart
epou, line 1: I think it's more scalable to call this user_provider an save the User instance, since in the future this provider can be used in other views. For this case, you can implement a user_score getter and return the score from the user.

DigitalSeneca: totally agree, I've refactored

lib/providers/user_provider.dart
```diff
@@ -0,0 +1,22 @@
...
+class UserProvider extends ChangeNotifier {
+  User? _user;
+  User? get user => _user;
+
+  int get userScore => _user?.score.value ?? 0;
+
+  Future<void> fetchUser() async {
+    try {
+      final response = await ApiSingleton.usersApi.retrieveMine();
+      if (response.data != null) {
```
epou, line 14: If `response.statusCode` is 200 and `response.data != null`, set `_user = response.data`, else set `_user =null`. Finally, notifyListeners()   (apply for all cases)

PR #385, opened 2025-06-27
Fixes using new api

DigitalSeneca:
Tested using a temporary button to execute the sending code 

| Screenshot 2025-06-27 at 17 51 46 | Screenshot 2025-06-27 at 17 52 24 |

### DIFF
--- a/lib/utils/BackgroundTracking.dart
+++ b/lib/utils/BackgroundTracking.dart
@@ -4,6 +4,7 @@ import 'dart:math';
 import 'package:battery_plus/battery_plus.dart';
 import 'package:flutter/material.dart';
 import 'package:geolocator/geolocator.dart';
+import 'package:mosquito_alert/mosquito_alert.dart';
 import 'package:mosquito_alert_app/api/api.dart';
 import 'package:permission_handler/permission_handler.dart';
 import 'package:shared_preferences/shared_preferences.dart';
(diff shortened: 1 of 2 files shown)


lib/utils/BackgroundTracking.dart
```diff
@@ -225,8 +226,19 @@ class BackgroundTracking {
       String trackingUuid = await BackgroundTracking._getTrackingUUID();
 
       // Send data to API
-      return await ApiSingleton().sendFixes(trackingUuid, position.latitude,
-          position.longitude, DateTime.now().toUtc(), batteryLevel);
+      FixesApi fixesApi = ApiSingleton.api.getFixesApi();
```
epou, line 229: I guess this is better to initialize in `api.dart` so in a future, you need to change the URL or any option you can manage everything in the same file.

DigitalSeneca: true, that will also avoid code duplication

lib/utils/BackgroundTracking.dart
```diff
@@ -225,8 +226,19 @@ class BackgroundTracking {
...
+        ..createdAt = DateTime.now().toUtc()
+        ..sentAt = DateTime.now().toUtc()
+        ..point.replace(FixLocationRequest((p) => p
+          ..latitude = position.latitude
+          ..longitude = position.longitude
+        ))
+        ..power = batteryLevel / 100.0
+      );
+      await fixesApi.create(fixRequest: fixRequest);
+      return true;
```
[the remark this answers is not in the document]
DigitalSeneca: I've updated it to return true only with status code 201

PR #405, opened 2025-07-04
Remove UUID creation on client side

### LINKED ISSUE #396: new api: wrong user initialization
User UUID creation should be on server side, not client side.
Therefore, code related to UUID creation on client side needs to be removed and replaced by the appropriate API call

Related to https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/381#discussion_r2182542668

DigitalSeneca:
fixes #396

### DIFF
--- a/lib/utils/Utils.dart
+++ b/lib/utils/Utils.dart
@@ -43,7 +43,6 @@ class Utils {
   static Map<String, dynamic> initializedCheckData = {
     'userCreated': {
       'created': false,
-      'required': true,
     },
     'firebase': false, // Whether firebase got initialized
   };


lib/pages/main/drawer_and_header.dart
```diff
@@ -247,63 +247,52 @@ class _MainVCState extends State<MainVC> {
...
-            return Text('Error: ${snapshot.error}');
-          }
-
-          return Row(
-            children: [
-              Text(
-                'ID: ',
+    return Consumer<UserProvider>(
+      builder: (context, userProvider, _) {
+        final uuid = userProvider.userUuid;
```
epou, line 252:
```suggestion
        final uuid = userProvider.user.uuid ?? '';
```

lib/utils/UserManager.dart
```diff
@@ -28,12 +27,9 @@ class UserManager {
       );
 
       await prefs.setBool('firstTime', true);
-      var uuid = Uuid().v4();
-      await prefs.setString('uuid', uuid);
```
epou, line 32: Nice! But then you should stop using `UserManager.getUUID()` in the other parts of code, in favor of the use of the provider: `userProvider.user.uuid`

lib/utils/Utils.dart
```diff
@@ -432,14 +431,9 @@ class Utils {
   }
 
   static Future<void> checkForUnfetchedData() async {
-    SharedPreferences prefs;
-    // if (userCreated["required"] && !userCreated["created"]) {
     final Map<String, bool> userCreated = initializedCheckData['userCreated'];
     if (userCreated['required']! && !userCreated['created']!) {
```
epou, line 435: (also remove the 'required' on the initalization of this dict)_
```suggestion
    if (!userCreated['created']!) {
```

lib/api/api.dart
```diff
@@ -74,7 +76,7 @@ class ApiSingleton {
     }
   }
 
-  Future<dynamic> createUser(String? uuid) async {
+  Future<dynamic> createUser() async {
```
epou, line 79: Having now the class UserProvider, I think it'll ease your development if you move this function there (also loginJwt). So as soon as the user is created, you can use the response to save it on the `_user` variable of the provider.

lib/utils/UserManager.dart
DigitalSeneca: I replaced the getUUID from SharedPreferences in favor of provider (it modifies many files on the same commit)
https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/pull/405/commits/e309e7967108d82c27eaf5307a025e6b9a62f6c8

lib/providers/user_provider.dart
```diff
@@ -21,4 +26,82 @@ class UserProvider extends ChangeNotifier {
...
+      if (registeredGuest.data?.username != null) {
+        final newApiUser = registeredGuest.data!.username;
+        await UserManager.setUser(newApiUser, guestPassword);
+        final success = await loginJwt(newApiUser, guestPassword);
+        if (success) {
+          await fetchUser();
+          return true;
+        }
+      }
+      return false;
```
epou, line 66:
```suggestion
        final username = response.data?.username;
        if (username == null) return false;

        await UserManager.setUser(username, guestPassword);

        if (!await loginJwt(username, guestPassword)) return false;

        await fetchUser();
       return true;
```

epou (approved):

PR #425, opened 2025-07-23
Added ApiService provider + improve auth flow

epou:


lib/pages/main/drawer_and_header.dart
```diff
@@ -86,15 +88,43 @@ class _MainVCState extends State<MainVC> {
...
+      // Create a guest user
+      password = Utils.getRandomPassword(10);
+      final GuestRegistration? guestRegistration =
+          await authProvider.createGuestUser(password: password);
+      // If guest user creation was successful, set the username to the guest username
+      if (guestRegistration != null) {
+        username = guestRegistration.username;
+      } else {
+        throw Exception("Failed to create guest user.");
+      }
```
DigitalSeneca, line 113:
```
      if (guestRegistration == null) {
        throw Exception("Failed to create guest user.");
      } 

      username = guestRegistration.username;
```

lib/providers/auth_provider.dart
```diff
@@ -0,0 +1,125 @@
+import 'package:device_info_plus/device_info_plus.dart';
+import 'package:dio/dio.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_secure_storage/flutter_secure_storage.dart';
+import 'package:mosquito_alert/mosquito_alert.dart';
+
+class AuthProvider with ChangeNotifier {
+  final FlutterSecureStorage _storage = const FlutterSecureStorage();
+  late final AuthApi authApi;
```
DigitalSeneca, line 9: If we are changing the way to invoke the api, we should also remove from api.dart to avoid the duplication
https://github.com/Mosquito-Alert/Mosquito-Alert-Mobile-App/blob/1f98200b53766847bd1ef6dcad399089602b1dc4/lib/api/api.dart#L20-L23

epou: The idea is to remove /lib/api/api.dart in favor of /lib/services/api_service.dart

PR #442, opened 2025-08-05
New api notifications fix

epou:


lib/pages/notification_pages/notifications_page.dart
```diff
@@ -67,20 +86,10 @@ class _NotificationsPageState extends State<NotificationsPage> {
...
-        }
-      }
-    } catch (e) {
-      print(e);
-    }
+  String _parseHtmlString(String htmlString) {
+    final document = parse(htmlString);
+    final String parsedString = document.body?.text ?? '';
+    return parsedString.trim();
   }
```
DigitalSeneca, line 84: Inside the try
```
    final notifId = widget.notificationId;
    if (notifId != null && notifId.isNotEmpty) {
      final notif = _pagingController.items?.firstWhere(
        (n) => notifId == n.id.toString(),
        orElse: () => null,
      );
      if (notif != null) {
        _infoBottomSheet(context, notif);
      }
    }
```

DigitalSeneca (approved):
I've tested it and looks pretty good

lib/pages/notification_pages/notifications_page.dart
epou: I don't understand this suggestion correctly. The function `_checkOpenNotification` has been removed in favor of the code added in the initState (see lines 56-65

DigitalSeneca: sorry I saw the diff incorrectly

PR #448, opened 2025-08-08
Refactor biting_form: Replace hardcoded values with enums

DigitalSeneca:


lib/pages/forms_pages/components/biting_form.dart
```diff
@@ -7,6 +7,39 @@ import 'package:mosquito_alert_app/utils/MyLocalizations.dart';
 import 'package:mosquito_alert_app/utils/Utils.dart';
 import 'package:mosquito_alert_app/utils/style.dart';
 
+enum BodyPart {
+  head(21),
+  left_arm(22),
```
epou, line 12: The convention in Dart is to use `lowerCamelCase` (affects to other enums too)

lib/pages/forms_pages/components/biting_form.dart
```diff
@@ -7,6 +7,39 @@ import 'package:mosquito_alert_app/utils/MyLocalizations.dart';
...
+  final int id;
+  const BodyPart(this.id);
+}
+
+enum BiteQuestion {
+  how_many(1),
+  bitten_body_part(2),
+  what_time(3),
+  vehicle_or_building(4),
+  when(5),
```
epou, line 27: 'when' looks like a reserved term for Dart

PR #465, opened 2025-08-21
BiteReportPage: Remove Campaign related code

DigitalSeneca:


lib/pages/forms_pages/biting_report_page.dart
```diff
@@ -175,42 +172,6 @@ class _BitingReportPageState extends State<BitingReportPage> {
       return;
     }
```
epou, line 174: shouldn't you call `showSuccess()` here? 🤔

PR #581, opened 2025-10-24
Bug fix: My reports Bites showing incorrect location

### LINKED ISSUE #560: My reports shows incorrect location
It shows something like "Location: Head" but it should be "Location: Barcelona"

Only for Europe: Backend returns location
Also available: getCityFromCoords

DigitalSeneca:
fixes #560 

Screenshot 2025-10-24 at 11 40 49
Screenshot 2025-10-24 at 11 36 01

### DIFF
--- a/lib/pages/my_reports_pages/components/reports_list_bites.dart
+++ b/lib/pages/my_reports_pages/components/reports_list_bites.dart
@@ -59,7 +59,7 @@ class _ReportsListBitesState extends State<ReportsListBites> {
   }
 
   String _formatCreationTime(Bite report) {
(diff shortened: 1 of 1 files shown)


lib/pages/my_reports_pages/components/reports_list_bites.dart
```diff
@@ -62,32 +62,23 @@ class _ReportsListBitesState extends State<ReportsListBites> {
     return DateFormat('yyyy-MM-dd HH:mm').format(report.createdAt.toLocal());
```
epou: createdAtLocal

PR #587, opened 2025-10-27
Migrate old auth system to new API auth

### LINKED ISSUE #585: Adapt old auth system to new API

epou:
Fixes #585

### DIFF
--- a/lib/pages/main/drawer_and_header.dart
+++ b/lib/pages/main/drawer_and_header.dart
@@ -137,7 +137,7 @@ class _MainVCState extends State<MainVC>
         SnackBar(
           content: Text(message),
           backgroundColor: Colors.red,
-          duration: Duration(seconds: 3),
+          duration: const Duration(seconds: 3),
         ),
       );
     }


lib/pages/main/drawer_and_header.dart
```diff
@@ -140,6 +154,22 @@ class _MainVCState extends State<MainVC>
...
+    final String? oldUsername = prefs.getString('uuid');
+    if (oldUsername != null) {
+      try {
+        await authProvider.login(
+            username: oldUsername, password: Env.oldPassword);
+        await authProvider.changePassword(
+            password: Utils.getRandomPassword(10));
+        await prefs.remove('uuid');
+      } catch (e) {
+        // TODO: If error is 5xx we should retry later.
```
DigitalSeneca: later when? In the next app launch or in X time within the same launch?

epou: After some minutes or in the next app launch. The idea would be to show the user a dialog informing that 'our servers are down, please try again in a few minutes',  or something like this.

PR #619, opened 2025-11-06
Remove unsude packages from pubspec

epou:


DigitalSeneca (approved):
nice! I had the vscode extension "Find Unused Dart Files & Assets" but this one seems better

PR #624, opened 2025-11-06
Refactor getPackageInfo

epou:


lib/pages/main/drawer_and_header.dart
```diff
@@ -373,7 +372,7 @@ class _MainVCState extends State<MainVC>
                   padding: const EdgeInsets.only(bottom: 20.0),
                   child: Text(
                     packageInfo != null
-                        ? 'version ${packageInfo.version} (build ${packageInfo.buildNumber})'
+                        ? 'version ${packageInfo!.version} (build ${packageInfo!.buildNumber})'
```
DigitalSeneca: do we want to show "version null (build null)" for a brief moment while the async method getPackageInfo loads the info instead of awaiting it?

Issue #479, opened 2025-09-08
Report buttons get stuck when there is no or low internet connectivity


JohnPalmer:
**Description**
When the user's phone has no data connection or an intermittent/slow data connection (as is the case, e.g., for many users in Bangladesh, if they click on any of the reporting buttons from the home screen, nothing happens, or at least nothing happens for a long time. As a result people assume the app is not working and they stop using it.

**To Reproduce**
1. Turn off your phone's data and wifi connections and open the app.
2. Click on any of the three report buttons (adult, site, or bite)
3. See how long it takes to open the report form.

**Expected behavior**
The app should be fully functional offline and should then sync with the server whenever connectivity is restored.

**Screenshots**
NA

**Smartphone (please complete the following information):**
 - Device: iPhone SE
 - OS: iOS18.6.2
 - Version 4.1.3 (2900)

**Additional context**
I have seen this happen on other phones as well, both Android and iOS.

PR #626, opened 2025-11-06
Implement retry on home page

### LINKED ISSUE #432: New API: When auth api request fails, there is an infinite loading circle on HomePage
**Describe the bug**
If the auth api request fails (ie timeout), then the HomePage is infinitely loading and the app is unusable: User does not have a UUID, user can't create reports, etc

**Screenshots**
Image

**Output**
```
I/flutter ( 1875): Error creating user: DioException [connection timeout]: The request connection took longer than 0:00:05.000000 and it was aborted. To get rid of this exception, try raising the RequestOptions.connectTimeout above the duration of 0:00:05.000000 or improve the response time of the server.
E/flutter ( 1875): [ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Exception: Failed to create guest user.
E/flutter ( 1875): #0      _MainVCState.initAuthStatus (package:mosquito_alert_app/pages/main/drawer_and_header.dart:118:9)
E/flutter ( 1875): <asynchronous suspension>
E/flutter ( 1875): #1      _MainVCState._startAsyncTasks (package:mosquito_alert_app/pages/main/drawer_and_header.dart:59:5)
E/flutter ( 1875): <asynchronous suspension>
E/flutter ( 1875):
```

**Possible solution**
- Increase connectTimeout (That's just a hack)
- Add some kind of retry (Better option)

epou:
Fixes #432 #479

### DIFF
--- a/lib/pages/main/drawer_and_header.dart
+++ b/lib/pages/main/drawer_and_header.dart
@@ -116,6 +116,7 @@ class _MainVCState extends State<MainVC>
       print('Failed to fetch notification count: $e');
       debugPrintStack(stackTrace: stackTrace);
     }
+    if (!mounted) return;
     setState(() {
       unreadNotifications = count;
     });


integration_test/background_tracking_easily_disabled_test.dart
```diff
@@ -52,11 +51,6 @@ void main() {
       expect(rejectBtn, findsOne);
       await tester.ensureVisible(rejectBtn);
       await tester.tap(rejectBtn);
-
-      // Home page
-      final homePageButtons = find.byType(CustomCard);
-      await waitForWidget(tester, homePageButtons);
-      expect(homePageButtons, findsNWidgets(4));
```
DigitalSeneca: It would be nice to keep the test otherwise there is no assertion after the user taps the rejectBtn and the test becomes useless.

If the workflow has changed, maybe we can change to expect the LoadingSpinner or other widget instead of the CustomCard

DigitalSeneca (commented):
I just tried in my physical phone where I got the DioException on main branch. Then I changed to this branch and it worked

PR #631, opened 2025-11-10
Add space between map and bottom of the screen

DigitalSeneca:
| Screenshot 2025-11-10 at 11 37 32 | Screenshot 2025-11-10 at 11 37 11 |


DigitalSeneca:
With the SafeArea it has a slightly different behavior.
I would think the SizedBox has a more natural behavior and shows the user they can scroll down, but we 

| Safe area | Sized box |
| safearea | sizedbox |

epou:
I see.... So then we can proceed with SizedBox

PR #635, opened 2025-11-10
Add new translations

epou:


DigitalSeneca (changes requested):
Some translations use the informal version while some use the formal. They should be unified in the preferred one. I just took a glance at Spanish and German

PR #673, opened 2025-12-22
Offline first

### LINKED ISSUE #95: Improve offline mode
Use:
https://pub.dev/documentation/http/latest/retry/RetryClient-class.html
https://pub.dev/packages/connectivity_plus

Example:
```dart
import 'dart:convert';
import 'package:http/http.dart' as http;
import 'package:http/retry.dart';
import 'package:connectivity/connectivity.dart';

class ApiHandler {
  static const String apiUrl = 'your_api_endpoint_here';
  static const int maxRetries = 5;
  static const Duration retryDelay = Duration(minutes: 1);

  static final client = RetryClient(
    http.Client(),
    retries: maxRetries,
    when: (http.BaseResponse response) {
      return response.statusCode >= 500; // Retry on 5xx status codes
    },
    delay: retryDelay,
  );

  static Future<dynamic> postFormData(Map<String, dynamic> formData) async {
    try {
      final connectivityResult = await Connectivity().checkConnectivity();
      if (connectivityResult == ConnectivityResult.none) {
        // No internet connection, retry until there's internet
        await _waitForInternetConnection();
      }

      final response = await client.post(Uri.parse(apiUrl), body: formData);
      
      if (response.statusCode == 200) {
        // API call successful, return response data
        return jsonDecode(response.body);
      } else {
        // Maximum retries reached, return error message
        return 'Max retries reached. Unable to reach the server.';
      }
    } catch (e) {
      // Handle other exceptions
      return 'An error occurred: $e';
    }
  }

  static Future<void> _waitForInternetConnection() async {
    while (true) {
      final connectivityResult = await Connectivity().checkConnectivity();
      if (connectivityResult != ConnectivityResult.none) {
        // Internet connection is available, break the loop
        break;
      }
      // Wait for a short duration before checking again
      await Future.delayed(Duration(seconds: 2));
    }
  }
}

```

### LINKED ISSUE #584: Implent async creation (offline)
Using local queue + list of pending reports.

Example: outbox from Gmail

Ideas:
- https://code.zeba.academy/offline-apps-flutter-syncing-data-handling-connectivity-issues/
- https://medium.com/@dhruvmanavadaria/building-offline-auto-sync-in-flutter-with-background-services-using-workmanager-13f5bc94023d
- https://knowledgebase.inheritxdev.in/syncing-data-offline-in-flutter-a-comprehensive-guide/
- https://dev.to/anurag_dev/implementing-offline-first-architecture-in-flutter-part-2-building-sync-mechanisms-and-handling-4mb1
- https://pub.dev/packages/brick_offline_first/install

epou:


JohnPalmer:
See the two pull requests I have made to fix minor issues I encountered on this branch when testing on iPhone. In addition:

1. I am not sure I would use the red "offline  mode" banner in the production version. I would rather users don't need to think about whether it is online or offline and that it simply work. That said, if you think it is important to show the they are offline, maybe just don't make it red, as they looks like an error more than normal functioning.

2. All of the reports I have been trying to send over the past hours end up with an icon of a crossed out cloud when I look at them in My reports. This despite my phone being connect to wifi and to a data connection and despite waiting in the hope they will sync. Please check if they are in fact not syncing (as opposed to the issue being really just about the app displaying an icon art the wrong time or the wrong icon.

epou:
Hey John,

I’ve slightly adapted the PR you opened and have already merged those changes into this branch.

1. Offline mode -> Good catch. I've changed the color from red to blue (blue aligns better with an informational state rather than a danger/error message)
2. When the app is offline, reports are saved locally. That’s why we show the crossed-cloud icon (it indicates the data isn’t coming from the “cloud” -> our platform). Once the device is back online, the code automatically attempts to post those locally stored reports to the API. I've just tested this flow on the emulator and it worked as expected.

I can also see 3 adult mosquito reports on the dev server from an iOS device last night (I’m guessing that was you)
- uKmI
- hwdS
- pmyk

PR #700, opened 2026-04-22
Fix silent deletion of offline reports on sync failure

JohnPalmer:
When an offline-created report's POST failed with a 4xx (or any non-DioException thrown inside sendCreateToApi, e.g. a missing photo file), the create action in ReportRepository caught the error and returned normally. OutboxMixin.execute() then treated the attempt as a success and deleted the local item from itemBox, even though the report had never reached the server. The report disappeared from 'My Reports' after the next refresh.

Two changes:

1. report_repository.dart: remove the blanket catch in the create outbox action so exceptions propagate. execute()'s catch block then reschedules the task as intended.

2. outbox_mixin.dart: move outbox.remove() to after a successful task.run(). Previously the outbox entry was removed before the task ran, so any failure path that couldn't reschedule cleanly would orphan the item.

### DIFF
--- a/lib/core/outbox/outbox_mixin.dart
+++ b/lib/core/outbox/outbox_mixin.dart
@@ -29,16 +29,20 @@ mixin OutboxMixin<
   /// Repository must implement method dispatcher
   Future<void> execute(OutboxTask task) async {
     final item = task.item;
-    await outbox.remove(item.id);
     try {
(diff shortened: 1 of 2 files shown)


JohnPalmer:
Eps - I am testing this now on my phone and it has bugs, Closing pull request.

PR #702, opened 2026-04-22
Fix offline report sync: prevent silent data loss and photo-cast crash

JohnPalmer:
__NOTE: All changes in this PR were done by Claude Opus 4.7. They have been tested briefly on iPhone SE via USB cable and seem to work; more tests coming__

## Summary

Fixes three bugs in the offline/sync path that together caused offline-created reports to disappear from the user's device (and never reach the server) under intermittent conditions. The first bug was silently masking the second; the third affected iOS gallery uploads specifically.

## Motivation / reported symptom

While testing on iPhone: a report created while offline appeared in **My Reports** marked as unsynced. After coming back online and pulling to refresh, the report sometimes correctly moved to synced, but occasionally **vanished entirely** — not on the device, not on the server, with no error shown to the user. Separately, reports with photos picked from the iOS gallery could get stuck in the outbox indefinitely.

## Root cause

Three distinct issues in the sync pipeline combined to produce this behaviour.

### 1. Silent deletion on any sync error

In `ReportRepository.buildOutboxTaskFromItem` the create action wrapped `_createApiOrLocal` in a blanket `catch (_) { break; }`. Any exception thrown by the POST (4xx response, a non-`DioException` such as a missing photo file, serializer/type errors, etc.) was swallowed and the action returned normally. `OutboxMixin.execute()` then treated that as success and deleted the local record from `itemBox`, even though the report had never reached the server.

Compounding this, `execute()` removed the outbox entry *before* running the task, so any failure path that couldn't reschedule cleanly would orphan the record.

### 2. `List<BasePhoto>` → `List<BaseUploadPhoto>` cast crash

Once the silent-deletion was fixed, testing exposed a second bug hiding behind it. `OutboxMixin.syncRepository()` calls `buildCreateRequestFromItem` on every queued report; for Observations and Breeding Sites this goes through `*CreateRequest.fromModel`, which did:

```dart
observation.photos as List<BaseUploadPhoto>
```

The field is declared `List<BasePhoto>?`, and Dart's generic collections are not covariant, so this throws `CastList<dynamic, BasePhoto> is not a subtype of List<BaseUploadPhoto>` the very first time sync runs — aborting `syncRepository()` before any POST, on every reconnection, forever. Online creates aren't affected because they bypass this path.

Once #1 was fixed, queued reports survived but got stuck indefinitely because of #2.

### 3. iOS gallery photos stuck in the outbox as HEIC-labelled-as-JPEG

Testing after #1 and #2 exposed a third failure mode on iOS gallery picks. `compressAndAddToSelectedImages` in `camera_with_gallery.dart` used `FlutterImageCompress.compressWithFile` as its only re-encode path, and when that returned `null` — which it does intermittently for certain HEIC/HEIF inputs (HDR, live-photo variants, some ICC profiles) — it fell back to pushing the raw file bytes into the upload list. Those bytes are HEIC, but the upload payload labels them `image/jpeg`. The server rejects the mislabeled payload, the report never syncs, and the outbox entry lives forever. No error was shown to the user.

## Changes

- **report_repository.dart** — remove the blanket `catch (_) { break; }` in the create outbox action so exceptions propagate. `execute()` will then reschedule via its existing error handling instead of treating a failed POST as success.
- **outbox_mixin.dart** — move `outbox.remove(item.id)` to *after* a successful `task.run()`, so a failing task never leaves a record orphaned between the outbox and `itemBox`. The reschedule path is idempotent (`_box.put` keyed by id), so this doesn't create duplicates.
- **observation_report_request.dart** and **breeding_site_report_request.dart** — replace the unsafe list cast with `photos?.whereType<BaseUploadPhoto>().toList() ?? []`, which filters element-wise and returns a correctly-typed list.
- **camera_with_gallery.dart** — when `compressWithFile` returns `null`, fall back to `FlutterImageCompress.compressWithList` at the *same* settings (4K long edge, quality 98, `keepExif`, `autoCorrectionAngle`). `compressWithList` routes through native ImageIO on iOS and handles HEIC/HEIF reliably. If both paths fail, the photo is dropped (never uploaded as raw bytes) and the user is told via a localized SnackBar so they can retake or reselect. A `failedImageCount` counter backs the notification; camera capture no longer pops back on failure, so the user can retry without losing previously-accepted photos in the session. The long-standing JPEG encoding settings (quality 98 / 4K / EXIF preserved) are unchanged.
- **en_US.json** — adds `photo_processing_failed` and `photo_processing_failed_plural` for the new user-visible error.

## Testing

Verified on an iPhone via `fvm flutter run --flavor devtf` and a TestFlight build:

- Creating an offline report and coming back online now successfully syncs the report to the server, where previously it crashed `syncRepository` with a type cast error.
- Simulated sync failures (via a proxy returning 4xx on the create endpoint) no longer silently delete the report; it remains visible as unsynced and is retried on the next connectivity transition.
- iOS gallery picks that previously stuck the outbox now upload successfully; photos that fail both compression paths are dropped and surfaced to the user instead of being silently queued.

## Known follow-ups (not in this PR)

Discovered during this investigation but intentionally deferred:

- Sync is only triggered on connectivity-state *transitions* and by the Workmanager background task. Pull-to-refresh on **My Reports** does not call `syncRepository()`, so a user who stays online doesn't get an automatic retry for a previously-stuck item. Worth adding as a small follow-up PR.
- The "no data connection" warning when starting an offline report flow is a pre-existing UX issue and unrelated to these fixes.
- **Server-side HEIC acceptance** would eliminate the entire class of iOS gallery bug at no quality cost and reduce upload sizes. Worth a conversation with the backend team.
- **Revisiting the 4K / quality-98 JPEG target** with the classification team is a separate scientific-fidelity decision and explicitly out of scope here; this PR preserves the existing settings unchanged.
- Broader hardening opportunities flagged in the review (centralized logging, typed exception hierarchy, response-schema validation, outbox/repository test coverage) are good candidates for separate PRs.

## Risk / backward compatibility

- No schema changes, no Hive migrations, no API changes.
- Behaviour change is strictly beneficial for the failure path: failed sync attempts are now retried rather than silently consumed, and malformed photo payloads are no longer queued.
- `whereType` is strictly safer than the previous cast; any previous code path where the old cast *didn't* throw would also produce the same list via `whereType`.
- The gallery compression chain is additive: the fast path is unchanged, and the new fallback uses identical quality settings. The only behavioural difference on failure is that we drop + notify instead of uploading bytes that the server rejects anyway.

### DIFF
--- a/assets/language/en_US.json
+++ b/assets/language/en_US.json
@@ -69,6 +69,8 @@
     "breeding_report_title": "Report a breeding site",
     "not_a_mosquito": "It is not a mosquito",
     "photo_required_alert": "You must attach at least one photo",
+    "photo_processing_failed": "Could not process that photo. Please try another photo or retake it.",
+    "photo_processing_failed_plural": "Could not process {count} of the selected photos. Please try different photos.",
     "bs_info_adult_title": "Attach a photo",
     "bs_info_adult_title_optional": "If you can, attach a photo",
     "continue_without_photo": "Continue without photo",
--- a/lib/core/outbox/outbox_mixin.dart
+++ b/lib/core/outbox/outbox_mixin.dart
@@ -29,16 +29,20 @@ mixin OutboxMixin<
   /// Repository must implement method dispatcher
   Future<void> execute(OutboxTask task) async {
     final item = task.item;
-    await outbox.remove(item.id);
     try {
       await task.run();
     } catch (error) {
-      // Only runs if task.run() fails
+      // Only runs if task.run() fails. The outbox entry is left in place so
+      // the task can be retried on the next sync cycle.
       await schedule(task, runNow: false);
       return; // stop further execution
     }
 
-    // Runs only if task.run() succeeded
+    // Only remove the outbox entry (and associated local item) after the
+    // task has successfully completed. Removing it beforehand risked losing
+    // the record entirely if the action threw after the remove but before
+    // reschedule could persist it.
+    await outbox.remove(item.id);
     if (item.operation == OutBoxOperation.create) {
       final request = createRequestFactory(item.payload);
       await itemBox.delete(request.localId);
--- a/lib/features/breeding_sites/data/models/breeding_site_report_request.dart
+++ b/lib/features/breeding_sites/data/models/breeding_site_report_request.dart
@@ -44,9 +44,10 @@ class BreedingSiteCreateRequest extends BaseCreateReportWithPhotosRequest {
             breedingSite.location.source_.name,
           ),
       ),
-      photos: breedingSite.photos != null
-          ? breedingSite.photos as List<BaseUploadPhoto>
-          : [],
+      // See ObservationCreateRequest.fromModel for the rationale: a
+      // `List<BasePhoto>` cannot be cast to `List<BaseUploadPhoto>`; filter
+      // element-wise instead.
+      photos: breedingSite.photos?.whereType<BaseUploadPhoto>().toList() ?? [],
       siteType: breedingSite.siteType,
       hasWater: breedingSite.hasWater,
       inPublicArea: breedingSite.inPublicArea,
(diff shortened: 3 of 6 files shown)


JohnPalmer:
This is still not working to send photos from Gallery. Closing PR until fixed.

PR #718, opened 2026-04-29
Added missing keychain-access-groups

epou:
Missing according to the documentation: https://pub.dev/packages/flutter_secure_storage#ios

Please see: https://github.com/juliansteenbakker/flutter_secure_storage/blob/v10.0.0/README.md#macos--ios


JohnPalmer:
I just deleted my previous comment. I have now tested the branch on iPhone SE and it it appears to work fine. We can go ahead and merge this.

PR #730, opened 2026-05-13
Migrate onboarding old system on init

epou:


epou:
No longer needed. See: https://github.com/Mosquito-Alert/mosquito_alert_mobile_app/pull/731

PR #738, opened 2026-05-18
Self-heal auth when stored credentials are rejected

JohnPalmer:
## Summary

Fixes the TestFlight regression where some testers see "phantom authenticated" 
state — the app behaves as if signed in (skips onboarding, no UUID shown in 
the drawer, report sync silently fails) because `hasCredentials()` returns 
true while the actual access token request is rejected by the server.

Root cause: on iOS the Keychain survives uninstall, so stale credentials 
(rotated passwords, staging-server tokens, deleted server-side accounts) 
can persist across reinstalls. The previous recovery logic only wiped 
credentials on HTTP 400 (refresh) or 400/401 (login), never re-attempted 
guest registration, and had no way to distinguish "first launch" from 
"reinstall on top of broken credentials".

## Changes

Three coordinated fixes in `lib/features/auth`:

1. **Track onboarding completion** with a SharedPreferences flag set on 
   every successful login and backfilled at init for existing installs 
   that already have keychain credentials. This lets us tell 
   "first-time user" apart from "returning user whose credentials broke".

2. **Broaden wipe criteria** in `AuthRepository.restoreSession`:
   - Refresh: wipe tokens on **any 4xx** (was: only 400).
   - Login: call `logout()` on **any 4xx** (was: only 400/401).
   - Non-`DioException` failures during login now also call `logout()`, 
     so the app cannot end up authenticated-in-memory but 
     credential-less-on-disk.
   - 5xx, timeouts, and connectivity errors continue to leave 
     credentials in place for later retry.

3. **Self-heal in `AuthProvider.restoreSession`**: when the repository 
   has no credentials but onboarding was previously completed, silently 
   create a fresh guest account so the user keeps using the app without 
   seeing onboarding again. If guest registration is offline-only, the 
   existing 15-minute outbox retry will eventually complete it.

## Testing

- `fvm flutter analyze lib/features/auth` → clean
- `fvm flutter test` → 24/24 pass
- Manual TestFlight verification pending on DevTF build.

## Risk / rollout

- No UI changes; self-heal is transparent.
- Users currently stuck in the phantom-authenticated state will, on next 
  app launch, automatically get a fresh guest account and resume 
  syncing — without losing the local outbox.
- No DB migration; the SharedPreferences flag is backfilled lazily.

### DIFF
--- a/lib/features/auth/data/auth_repository.dart
+++ b/lib/features/auth/data/auth_repository.dart
@@ -26,6 +26,13 @@ class AuthRepository with OutboxMixin<AuthUser, UserRegistrationRequest> {
   static const _deviceIdKey = 'device_id';
   static const _autoChangePasswordKey = 'auto_change_password';
 
+  /// SharedPreferences flag indicating the user has, at some point in the
+  /// past, successfully obtained credentials on this install. Used to decide
+  /// whether to silently self-heal (re-create a guest account) when stored
+  /// credentials are later rejected by the server, instead of bouncing the
+  /// user back to onboarding.
+  static const _onboardingCompletedKey = 'auth_onboarding_completed';
+
   late final AuthApi _authApi;
 
   final Future<Device> Function()? getCurrentDevice;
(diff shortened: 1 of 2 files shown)


epou:
I think it’s better to keep the current behavior and not revert the changes introduced in PR #731.

The main reason is that we only delete the stored username and password when the API explicitly indicates that the credentials are no longer valid (HTTP 400 or 401). In that situation, we know that the credentials stored on the device no longer match the server database, so the only safe option is to start from scratch.

With the current approach, the user is taken through the onboarding flow again, which makes it clear that their previous local session is no longer valid and that a new guest account will be created. This provides a much more transparent experience. If we automatically created a new guest user in the background instead, users might continue using the app without realizing that they are now associated with a different account. This could lead to confusion and support requests from users who believe they have “lost” their reports or can no longer access data associated with their previous account.

In any case, this is an edge case that only occurs when the server explicitly rejects the stored credentials, which is something abnormal. Since there is no reliable way to recover the original account automatically, restarting the onboarding process is the safest and most predictable behavior. This is better aligned with potential future login/logout functionality, where invalid credentials would naturally require the user to authenticate again rather than silently creating a new account.

I would also suggest taking a look at PR #737, which makes iOS start with an empty Keychain on every fresh installation, mimicking the behavior already implemented on Android. This ensures consistent behavior across both platforms and avoids platform-specific differences in how credentials are preserved between installations (if android user can not restore a session on reinstallation, iOS user shouldn't either). As a result, users will have the same experience regardless of whether they are using iOS or Android.

